Issue #6, opened 2018-06-04
ics/ical script support for multiple calendars


derekantrican:
From Google-Apps-Script-Library created by [derekantrican](https://github.com/derekantrican) : derekantrican/Google-Apps-Script-Library#10  

The argument here is to add support for multiple ics/ical calendars to be imported into the same calendar

agentd00nut:
See #9

p5hema2:
Simply Copy the Script

derekantrican:
@p5hema2 Making copies of the script is a possible temporary solution, but a PR is already being worked on for a final solution. You can see pull requests related to this issue at #36 and #9 (I'm waiting on approval of #31 to happen first before I start a new PR for this issue, because it will greatly refactor the code)

lbell:
+1 - I see the pull requests in progress, so don't want to muddy the waters by diving in head first. How can I help make this happen?

derekantrican:
@lbell We're working on #47 first and if you feel like you can contribute to that upgrade you can do pulls (or test, open issues, etc) to @jonas0b1011001 's repo here: https://github.com/jonas0b1011001/GAS-ICS-Sync-CalAPIv3

derekantrican:
### Closing as this is fixed by the latest release: [v5.0](https://github.com/derekantrican/GAS-ICS-Sync/releases/tag/5.0)

*If you find that there are still bugs, please comment on this issue and let us know*

PR #9, opened 2018-06-07
Multi calendar

agentd00nut:
Takes relatively little code to make this happen surprisingly.

Currently does nothing to alleviate the issues i've faced with going over API quotas or execution times

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -12,12 +12,15 @@
 // var targetCalendarName = "" // The name of the Google Calendar you want to add events to
 // var sourceCalendarURL = "" // The ics/ical url that you want to get events from
 
-var sourceCalendars={}
+
+var sourceCalendars={
+  CustomName:"ICS/ICAL URL"
+}
 
 
 // Currently global settings are applied to all sourceCalendars.  
 
-var howFrequent = 15; //What interval (minutes) to run this script on to check for new events
+var howFrequent = 30; //What interval (minutes) to run this script on to check for new events
 var addEventsToCalendar = true; //If you turn this to "false", you can check the log (View > Logs) to make sure your events are being read correctly before turning this on
 var modifyExistingEvents = true; // If you turn this to "false", any event in the feed that was modified after being added to the calendar will not update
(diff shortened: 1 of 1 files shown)


agentd00nut:
#6

Code.gs
```diff
@@ -12,12 +12,15 @@
 // var targetCalendarName = "" // The name of the Google Calendar you want to add events to
```
derekantrican: If we're going to deprecate `targetCalendarName` and `sourceCalendarURL`, let's just remove those options

Code.gs
```diff
@@ -12,12 +12,15 @@
...
 // var sourceCalendarURL = "" // The ics/ical url that you want to get events from
 
-var sourceCalendars={}
+
+var sourceCalendars={
+  CustomName:"ICS/ICAL URL"
+}
```
derekantrican: Rather than having `sourceCalendars` "prepopulated" with example data, let's leave it blank (like you did before with `var sourceCalendars = { };` ) and put a comment before explaining how it should be used (even use a block comment if you need to)

Code.gs
```diff
@@ -12,12 +12,15 @@
...
 
 
 // Currently global settings are applied to all sourceCalendars.  
 
-var howFrequent = 15; //What interval (minutes) to run this script on to check for new events
+var howFrequent = 30; //What interval (minutes) to run this script on to check for new events
```
derekantrican: Why is this being changed again? Leftover from testing?

Code.gs
```diff
@@ -156,6 +170,7 @@ function main(){
   //----------------------------------------------------------------
   
   if(addEventsToCalendar || removeEventsFromCalendar){
+    /* We might want to consider reducing how far back we reach for events to reduce API call usage... Also would make a good premium feature */
```
derekantrican: Good point, but let's take this comment out so we don't have comments about "possible premium features" in our code :)

derekantrican:
@agentd00nut Note that this branch will have to be redone since https://github.com/derekantrican/GAS-ICS-Sync/pull/20

agroszer:
any chance to get this up to speed?

Issue #11, opened 2018-06-13
Script will delete events not related to ICS file


derekantrican:
One of the beauties of this custom script is that it can sync to an already-existing calendar (where there are other events not related to the ICS/ICAL file). But this means that those events won't have a `FID` tag on them.

We need to fix this section of the program:

    var feedIndex = feedEventIds.indexOf( calendarEvents[i].getTag("FID") );
    if(removeEventsFromCalendar){
      if( feedIndex  == -1  ){
        Logger.log("    Deleting "+calendarEvents[i].getTitle());
        calendarEvents[i].deleteEvent();
      }
    }

The problem is that it is deleting events that may be on the calendar that are not related to the ICS/ICAL file.

agentd00nut:
Check the calendar for events that don't have an FID and don't have an "OriginalEvent" tag.  We add the OE tag to those events.

Then we do the FID comparison... except we only remove events with no FID that also don't have the OE tag. `feedIndex == -1 && calendarEvents[i].getTag("OE") == false` ( not sure what getTag returns on a miss... )


That way everything there during a first sync gets OE tag and will never be removed by the script.

Events from a source calendar that are deleted from the source will be removed as expected.

Events added to the destination calendar won't have an FID, so during the pass they'll get an OE tag.  So they won't be removed.

derekantrican:
This has now been added to the script at the regular public URL: https://script.google.com/d/1QeZFLSM1EkuFvYcryECI_xH-IZVe1-IxGRq_n6OoXp1CmVtSeTeigEx4/edit?usp=sharing

PR #13, opened 2018-06-20
Added check for null tag value #11

derekantrican:
Added a check for a null value of the `FID` tag on events. This way, we aren't deleting events that are on the calendar, but aren't part of the ICS/ICAL file


agentd00nut:
Events that exist before the initial sync will have a null FID value.
And events manually added after an initial sync will have a null FID value.

This properly avoids deleting those kinds of events, nice one!

PR #17, opened 2018-08-24
Added version checking

### LINKED ISSUE #12: Check for new version
Eventually, it would be nice to have a function that checks for a new version of the program from here (probably from the GitHub releases page?)

Something like this:

1. Every time the `main` function runs, it calls the `CheckForUpdate` function
2. `CheckForUpdate` will check the current version (a global variable in the code) against the most recent published version on GitHub releases page
3. If a newer version is found, we will check a [`ScriptProperty`](https://developers.google.com/apps-script/reference/properties/) to make sure we haven't found that before. If we haven't, we will email the user that there is a new version (and set that `ScriptProperty`). If we have, do nothing

derekantrican:
Resolves #12

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -1,77 +1,100 @@
-/* --------------- HOW TO INSTALL ---------------
+/*
+*=========================================
+*       INSTALLATION INSTRUCTIONS
+*=========================================
 *
 * 1) Click in the menu "File" > "Make a copy..." and make a copy to your Google Drive
-* 2) Changes lines 11-20 to be the settings that you want to use
+* 2) Changes lines 19-32 to be the settings that you want to use
 * 3) Click in the menu "Run" > "Run function" > "Install" and authorize the program
 *    (For steps to follow in authorization, see this video: https://youtu.be/_5k10maGtek?t=1m22s )
-* 4) To stop Script from running click in the menu "Edit" > "Current Project's Triggers".  Delete the running trigger.
+*
+*
+* **To stop Script from running click in the menu "Edit" > "Current Project's Triggers".  Delete the running trigger.
+*
+*=========================================
+*               SETTINGS
+*=========================================
 */
 
-// --------------- SETTINGS ---------------
+var targetCalendarName = "";           // The name of the Google Calendar you want to add events to
+var sourceCalendarURL = "";            // The ics/ical url that you want to get events from
 
-var targetCalendarName = "" // The name of the Google Calendar you want to add events to
-var sourceCalendarURL = "" // The ics/ical url that you want to get events from
+var howFrequent = 15;                  // What interval (minutes) to run this script on to check for new events
+var addEventsToCalendar = true;        // If you turn this to "false", you can check the log (View > Logs) to make sure your events are being read correctly before turning this on
+var modifyExistingEvents = true;       // If you turn this to "false", any event in the feed that was modified after being added to the calendar will not update
+var removeEventsFromCalendar = true;   // If you turn this to "true", any event in the calendar not found in the feed will be removed.
+var addAlerts = true;                  // Whether to add the ics/ical alerts as notifications on the Google Calendar events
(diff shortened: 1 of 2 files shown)


Code.gs
```diff
@@ -279,19 +312,26 @@ function ParseNotificationTime(notificationString){
...
+      var regex = RegExp("<a.*title=\"\\d\\.\\d\">","g");
+      var latestRelease = regex.exec(html)[0];
+      regex = RegExp("\"(\\d.\\d)\"", "g");
+      var latestVersion = Number(regex.exec(latestRelease)[1]);
+      
+      if (latestVersion > thisVersion){
+        if (email != ""){
+          GmailApp.sendEmail(email, "New version of GAS-ICS-Sync is available!", "There is a new version of \"GAS-ICS-Sync\". You can see the latest release here: https://github.com/derekantrican/GAS-ICS-Sync/releases");
+        
+          PropertiesService.getScriptProperties().setProperty("alertedForNewVersion", true);
```
Merlin2001, line 331: I see a problem here, as this script property is never reset (e.g. during `install()`) so I expect the update check to work _only once_ and then always fail because `alertedForNewVersion` stays true.

As a solution I would suggest to store the last _version number_ we alerted for, instead of only a boolean value. This way the user will be alerted again if a newer new version is available 😄 

Suggested new implementation:
```js
function checkForUpdate(){
  // No need to check if we can't alert anyway
  if (email == "")
    return;

  var lastAlertedVersion = PropertiesService.getScriptProperties().getProperty("alertedForNewVersion");
  try {
    var thisVersion = 5.4;
    var latestVersion = getLatestVersion();

    if (latestVersion > thisVersion && latestVersion != lastAlertedVersion){
      GmailApp.sendEmail(email,
        "New version of GAS-ICS-Sync is available!",
        "There is a new version of \"GAS-ICS-Sync\". You can see the latest release here: https://github.com/derekantrican/GAS-ICS-Sync/releases");

      PropertiesService.getScriptProperties().setProperty("alertedForNewVersion", latestVersion);
    }
  }
  catch (e){}

  function getLatestVersion(){
    var html = UrlFetchApp.fetch("https://github.com/derekantrican/GAS-ICS-Sync/releases");
    var regex = RegExp("<a.*title=\"\\d\\.\\d\">", "g");
    var latestRelease = regex.exec(html)[0];
    regex = RegExp("\"(\\d.\\d)\"", "g");
    return Number(regex.exec(latestRelease)[1]);
  }
}
```

Also, the comment for the `email` variable in Code.gs needs to be updated from:  
`// OPTIONAL: If "emailWhenAdded" is set to true, you will need to provide your email`
to  
`// OPTIONAL: If "emailSummary" is set to true or you want to receive notifications when a new version is available, you will need to provide your email address`

derekantrican: Yeah, this was added that way because users can run this script quite often (maybe every 5 minutes) and we don't want to start sending someone an email every 5 minutes that a new version is available.

If I understand correctly, what you're suggestion is the current workflow (eg user has version 5 and we alert them once that version 6 has come out) but with the addition that we alert for every new version (eg user has version 5, we alerted them for version 6, they still haven't upgraded and version 7 comes out and we alert them again)?

Merlin2001: Exactly, that's what my change is intended to achieve.

But I also just understood that a user will usually "upgrade" to the new version by copying the script from your account to theirs again (so they have a new project and `alertedForNewVersion` will be `false` for that new project.

So with the current code every user gets one update notification unless they actually update, and that's totally fine for me (I just thought that _even if_ I updated, the script property would stay true thus preventing all but the first ever notification. That was a misunderstanding of the update process on my side.)

The `email` comment "issue" still stands, though 😁

derekantrican: Yes, exactly - the variable is reset when the new script is copied (provided they use the "Copy to my Google Drive" option and don't just copy and replace the code)

What is "the `email` comment "issue""? I think I missed that

Merlin2001: I just went ahead and created #163 because a diff says more than a thousand words 😄

PR #19, opened 2019-01-26
Add organiser name to event summary

JoelBalmer:
Hi there, really like your script! I've been using it but sometimes an event is created and i don't know who it's from without investigating further. I thought it would be neat to include the name in the event summary so this code allows for that. Let me know if agree or would like me to change anything. Cheers!


derekantrican:
@JoelBalmer I am doing a massive overhaul to the program right now and trying to make sure I keep all current functionality. Would you mind sharing your ical/ics url so that I can make sure this "organizer" change is preserved?

JoelBalmer:
Sure thing, I've emailed you (potentially secret events, etc.)

Issue #25, opened 2019-02-06
Same events are not recognized


michaelcadilhac:
Hi there;

I've been hit by a steady flow of:

    You have been creating or deleting too many calendars or calendar events in a short time.

from Google, so I investigated a bit.  It seems that lots of events were modified because a spurious change was detected.  Two reasons for that:

1. **null vs empty string**

The test `e.getLocation() != fe.location` failed because one was `null` while the other was the empty string.  This is fixed by changing 

    event.location = vevent.getFirstPropertyValue('location');

to

    event.location = vevent.getFirstPropertyValue('location') || '';

(This also applies to `description`!)

2. **comparing dates**

Comparing dates should not be done with `==`, but using `getTime`.  The line used to compare time should be changed to:

    if(e.getStartTime().getTime() != fe.startTime.getTime() || e.getEndTime().getTime() != fe.endTime.getTime())

Cheers;
M.

derekantrican:
Wow, @michaelcadilhac - thank you for the research and input! If you turn this into a pull request, I'd be happy to approve it and add you as a contributor to the repository

johnli56789:
I have the latest script, and yet I still encounter this error when importing my ics calendar.

**_Error
You have been creating or deleting too many calendars or calendar events in a short time. Please try again later. at main(Code:190)_** 

Line 190 is this code, **_resultEvent.setTag("FID", event.id);_**

This continues to happen even though I manually imported the ics calendar first, so any new sync should just be for the new events, unless there is something else in my ics file that cause Google Calendar import to treat every events as new events. How can I confirm if that is the case? 

This ics file has around 1000 events. I also tested with another ics file with just 60 events and this worked. Is there a limit of events that can be imported at a single time? Thanks

derekantrican:
@johnli56789 You can see if you Google this error, most of the results will suggest using `Utilities.sleep` (which "pauses" the execution for a defined number of miliseconds). I would suggest adding `Utilities.sleep(500);` at the end of that `for` loop (after line 197). Please let me know if that works for you.

BearBritches:
Instead of the error previous error "....creating or deleting too many...", I'm getting a "Timed Out" error.

PR #26, opened 2019-02-06
Sanitize event properties before equality check.

michaelcadilhac:
Fixing #25 (and a small typo).

Cheers;
M.


derekantrican:
Thanks for doing that! I'll update the public script to the latest version

michaelcadilhac:
My pleasure, Derek.  Thanks for your work; that script certainly is a must for _lots_ of applications.

PR #28, opened 2019-02-15
Handling of all day events

derekantrican:
Adding handling of all day events

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -173,6 +173,8 @@ function main(){
           
         if (emailWhenAdded)
           GmailApp.sendEmail(email, "New Event Added", "New event added to calendar \"" + targetCalendarName + "\" at " + event.startTime);
+          
+        Utilities.sleep(250); //Solve some "Creating too many events in a short time" issues
       }
     }
   }
--- a/Code.gs
+++ b/Code.gs
@@ -174,8 +174,6 @@ function main(){
           
         if (emailWhenAdded)
           GmailApp.sendEmail(email, "New Event Added", "New event added to calendar \"" + targetCalendarName + "\" at " + event.startTime);
-          
-        Utilities.sleep(250); //Solve some "Creating too many events in a short time" issues
       }
     }
   }
--- a/Code.gs
+++ b/Code.gs
@@ -145,13 +145,25 @@ function main(){
     Logger.log("Checking " + events.length + " Events for creation")
     for each (var event in events){
       if (calendarFids.indexOf(event.id) == -1){
-        var resultEvent = targetCalendar.createEvent(event.title, 
-                                                     event.startTime,
-                                                     event.endTime,
-                                                     {
-                                                      location : event.location, 
-                                                      description : event.description
-                                                     });
+        var resultEvent;
+        if (event.isAllDay){
+          resultEvent = targetCalendar.createAllDayEvent(event.title, 
+                                                         event.startTime,
+                                                         event.endTime,
+                                                         {
+                                                           location : event.location, 
+                                                           description : event.description
+                                                         });
+        }
+        else{
+          resultEvent = targetCalendar.createEvent(event.title, 
+                                                   event.startTime,
+                                                   event.endTime,
+                                                   {
+                                                     location : event.location, 
+                                                     description : event.description
+                                                   });
+        }
         
         resultEvent.setTag("FID", event.id);
         Logger.log("   Created: " + event.id);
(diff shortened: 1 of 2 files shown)


Code.gs
```diff
@@ -96,6 +96,7 @@ function main(){
   if(targetCalendar == null){
       Logger.log("Creating Calendar: " + targetCalendarName);
       targetCalendar = CalendarApp.createCalendar(targetCalendarName);
+      targetCalendar.setSelected(true); //Sets the calendar as "shown" in the Google Calendar UI
```
michaelcadilhac: Isn't that imposing on the user?

Code.gs
```diff
@@ -145,13 +146,25 @@ function main(){
...
-        var resultEvent = targetCalendar.createEvent(event.title, 
-                                                     event.startTime,
-                                                     event.endTime,
-                                                     {
-                                                      location : event.location, 
-                                                      description : event.description
-                                                     });
+        var resultEvent;
+        if (event.isAllDay){
+          resultEvent = targetCalendar.createAllDayEvent(event.title,
```
michaelcadilhac: Where is `createAllDayEvent` implemented?

Code.gs
```diff
@@ -161,6 +174,8 @@ function main(){
           
         if (emailWhenAdded)
           GmailApp.sendEmail(email, "New Event Added", "New event added to calendar \"" + targetCalendarName + "\" at " + event.startTime);
+          
+        Utilities.sleep(250); //Solve some "Creating too many events in a short time" issues
```
michaelcadilhac: This seems unrelated to handling all day events, isn't it?

Code.gs
```diff
@@ -230,6 +245,9 @@ function ConvertToCustomEvent(vevent){
   var dtstart = vevent.getFirstPropertyValue('dtstart');
   var dtend = vevent.getFirstPropertyValue('dtend');
   
+  if (dtstart.isDate && dtend.isDate)
```
michaelcadilhac: How does that work?

Code.gs
derekantrican: Possibly, but I have a feeling we would get more complaints of "it's not creating any events" [lots of troubleshooting later...] "oh, the calendar just wasn't shown in Google Calendar"

Code.gs
derekantrican: Here: https://developers.google.com/apps-script/reference/calendar/calendar#createAllDayEvent(String,Date,Date,Object)

Code.gs
derekantrican: Sure. Do you think this should be separate, then?

Code.gs
derekantrican: Comes from here: http://mozilla-comm.github.io/ical.js/api/ICAL.Time.html (description: `If true, the instance represents a date (as opposed to a date-time)`). Basically, the `DTSTART` (or `DTEND`) object on the `VEVENT` can either be a date or datetime. You can see the discussion about that here: https://github.com/mozilla-comm/ical.js/issues/353

derekantrican:
@michaelcadilhac Did you have any input on my replies?

michaelcadilhac:
@derekantrican Aye, sorry for the lag, busy week.  Can I get back to you during the weekend?

derekantrican:
@michaelcadilhac Sure, no prob

Code.gs
michaelcadilhac: Well yes probably :-)

Code.gs
michaelcadilhac: Ah fantastic, thanks for digging that up.

Code.gs
derekantrican: Ok, this should be fixed now

michaelcadilhac:
Note that this fails to modify events that are turned into all day events.

derekantrican:
@michaelcadilhac Thanks for letting me know. Please add a new bug with any helpful information (eg see [this issue](https://github.com/derekantrican/GAS-ICS-Sync/issues/34) for good helpful information

boillodmanuel:
Hello,

I added support for modifying all day events on my branch. As our codes has diverged, I can't open a pull request, but feel free to look at my code and reuse what you want : https://github.com/boillodmanuel/GAS-ICS-Sync/commit/4f72d13f168493bb40704fe93efbc354226bf29a

As far as I know, the main difference between our codes is the way we handle recurring events.

derekantrican:
@boillodmanuel Good to know! I'll look at incorporating those changes. Currently I'm quite busy with other projects but hoping to get back into this one soon

PR #31, opened 2019-02-18
Recurrence

derekantrican:
Added handling of event recurrence

Still present shortcomings/issues:

- Cannot handle negative `BYMONTHDAY` (eg `-3` for events that repeat on the third-to-last day of the month). Submitted a bug for this to Google: https://issuetracker.google.com/issues/124579536
- No handling for `EXDATE` (recurrence exclusions). This is a known issue with ICAL.js (https://github.com/mozilla-comm/ical.js/issues/257) and we may be able to provide a workaround in the future (or just adopt [ical-expander](https://github.com/mifi/ical-expander) instead)


derekantrican:
This now needs to be updated for these changes: https://github.com/derekantrican/GAS-ICS-Sync/pull/28

derekantrican:
Final testing needs to be done to check modification and deleting of recurring events before this can be approved

derekantrican:
Ok, other than the shotcomings mentioned in the PR description (`BYMONTHDAY` & `EXDATE`) this should be ready to merge

derekantrican:
After doing testing, this looks almost perfect. Not parsing `EXDATE` is obviously a shortcoming.

One problem is that All Day events seem to be created on the wrong day:

image

This may be due either to these pending changes or the changes made in https://github.com/derekantrican/GAS-ICS-Sync/pull/28

Code.gs
```diff
@@ -178,46 +171,82 @@ function main(){
...
+          continue;
+          
+        var fe = fes[0];
+        
+        if (fe.recurrence != null){
+          var eSeries = calendarEvents[i].getEventSeries();
+          
+          //Since there is no CalendarEventSeries.getRecurrence() method we can use to check then we will always set the recurrence
+          //(which also means setting the time). This also solves the situation of a regular event becoming a recurring event.
+          eSeries.setRecurrence(fe.recurrence, fe.startTime, fe.endTime);
```
jonas0b1011001: setRecurrence does not delete the exisiting recurring events.
So if you change an event from Interval 2 days to interval 3 days you will end up with it repeating every 2 and every 3 days..Might have to delete the old Eventseries and freshly create the new one.

Code.gs
```diff
@@ -178,46 +171,82 @@ function main(){
...
+        }
+        else{
+          var e = calendarEvents[i];
+          
+          var occurences = calendarFids.filter(function(value){ return value === tagValue; }).length;
+          if (occurences > 1){
+            //Event used to be a recurring event, but is now a regular event. Delete the recurring series from
+            //the calendar and create the new event fresh
+            e.getEventSeries().deleteEventSeries();
+            CreateEvent(targetCalendar, fe);
```
jonas0b1011001: Have to set the ID of the new event

jonas0b1011001: The checks below should be the else case, no need to check for changes after freshly adding the new event

Code.gs
```diff
@@ -271,57 +304,60 @@ function ConvertToCustomEvent(vevent){
...
+    
+      resultEvent = targetCalendar.createAllDayEvent(customEvent.title, 
+                                                     customEvent.startTime,
+                                                     customEvent.endTime,
+                                                     {
+                                                       location : customEvent.location, 
+                                                       description : customEvent.description
+                                                     });
+                                                     
+      resultEvent = resultEvent.getEventSeries().setRecurrence(customEvent.recurrence,
```
jonas0b1011001: There's no need to set the start and end time for the EventSeries. Using setRecurrence(event, startDate) works just fine and will not turn AllDay Events to normal Events.

derekantrican:
Closing this PR because we'll be doing recurrence through the full Google Calendar API and the fork [here](https://github.com/jonas0b1011001/GAS-ICS-Sync-CalAPIv3) (which will then get pulled back into this repo).

However, a lot of the code refactoring changes that were done in this PR should still be implemented

Issue #40, opened 2019-04-11
End time cannot be same 


andytrine:
For one-day all-day events I get an error: Event start date must be before event end date. (line 172, file "Code")

Example:
[19-04-11 10:38:53:653 CEST] Title: New Year's Day
[19-04-11 10:38:53:654 CEST] Id: 8299f54d-117f-46c9-84f9-ae35fe12d7c7@vacationtracker.io
[19-04-11 10:38:53:654 CEST] Description: 
[19-04-11 10:38:53:654 CEST] Start: Tue Jan 01 2019 00:00:00 GMT-0800 (PST)
[19-04-11 10:38:53:655 CEST] End: Tue Jan 01 2019 00:00:00 GMT-0800 (PST)

derekantrican:
@andytrine if you are willing, please share with me your ICS file

andytrine:
@derekantrican sure - here is the ical link, hope that works:
https://api.vacationtrackerbot.com/calendar/jointrine/c87d5fa8-032c-42cd-8ae5-1ba04b1bf1d5/upcoming.ical

andytrine:
Also once it does an update synch on existing events they become "normal" events and not all-day events for some reason

andytrine:
Also when it synchs it makes duplicates of the same event?

jonas0b1011001:
> Also when it synchs it makes duplicates of the same event?

So did you get it to sync the ics you shared above? The event-ids used are actually not supported by Google Calendar.

jonas0b1011001:
As mentioned [here](https://github.com/derekantrican/GAS-ICS-Sync/issues/55#issuecomment-534003159) this is not a valid way to describe an one-day all-day event.

derekantrican:
@andytrine I would suggest contacting the developers at "vacationtrackerbot.com" and alert them of this issue. As @jonas0b1011001 mentioned, this is not a valid format for an "all-day" event according to the [official ical documentation](https://tools.ietf.org/html/rfc5545#section-3.8.2.2):

```
BEGIN:VEVENT
UID:2336d2bf-463b-4a23-afad-04b2a809d212@vacationtracker.io
SEQUENCE:0
DTSTAMP:20190925T131403Z
DTSTART;VALUE=DATE:20190101
DTEND;VALUE=DATE:20190101
X-MICROSOFT-CDO-ALLDAYEVENT:TRUE
[...]

Issue #38, opened 2019-03-28
Feature: Add Description and Title to Email Notifications 


p5hema2:
Add Description and Title to Email Notifications

derekantrican:
It looks like there was a regression here. It was fixed in https://github.com/jonas0b1011001/GAS-ICS-Sync-CalAPIv3/commit/1ddd0dc845cdb38943d43adf235622dce5f7d026 but the latest version only includes Titles in email notifications, not descriptions. We'll work on that

Issue #34, opened 2019-02-26
If Script's timezone is different than calendar, all day events are created on the wrong day


derekantrican:
### Setup:

**Calendar timezone:** Central time - Chicago
**Script timezone:** Pacific Time
**Test string ([wiki](https://github.com/derekantrican/GAS-ICS-Sync/wiki/Debugging-tips)):**

    var sourceCalendarString =
    "BEGIN:VCALENDAR\n"+
    "PRODID;X-RICAL-TZSOURCE=TZINFO:-//com.denhaven2/NONSGML ri_cal gem//EN\n"+
    "CALSCALE:GREGORIAN\n"+
    "VERSION:2.0\n"+
    "METHOD:PUBLISH\n"+
    "BEGIN:VTIMEZONE\n"+
    "TZID;X-RICAL-TZSOURCE=TZINFO:America/Chicago\n"+
    "BEGIN:STANDARD\n"+
    "DTSTART:20181104T020000\n"+
    "RDATE:20181104T020000\n"+
    "TZOFFSETFROM:-0500\n"+
    "TZOFFSETTO:-0600\n"+
    "TZNAME:CST\n"+
    "END:STANDARD\n"+
    "BEGIN:DAYLIGHT\n"+
    "DTSTART:20190310T020000\n"+
    "RDATE:20190310T020000\n"+
    "TZOFFSETFROM:-0600\n"+
    "TZOFFSETTO:-0500\n"+
    "TZNAME:CDT\n"+
    "END:DAYLIGHT\n"+
    "END:VTIMEZONE\n"+
    "BEGIN:VEVENT\n"+
    "DTEND;VALUE=DATE:20190408\n"+
    "DTSTART;VALUE=DATE:20190407\n"+
    "DTSTAMP;VALUE=DATE-TIME:20190221T204949Z\n"+
    "UID:event_448196497_ngin.com\n"+
    "DESCRIPTION:Tag(s): 14U Mascheri\nhttps://sportsengine.page.link/?al=spor\n"+
    " tsengine%3A%2F%2Fevent%2F448196497%3Fevent_start_time%3D2019-04-07T05%25\n"+
    " 3A00%253A00Z%26team_ids%255B%255D%3D2830426&apn=com.sportngin.android&ct\n"+
    " =iCalEventLink&ibi=com.ngin.sport.iphone&isi=499597400&link=http%3A%2F%2\n"+
[...]

Issue #29, opened 2019-02-17
Using vtimezone make script very long to execute


boillodmanuel:
Hi,
I figure out than using `vtimezone` make the script very long to execute. 
If I remove all code related to `vtimezone` in `Code.js`, execution time decrease from 120s to 2s.
The time is taken by `toJSDate()` method. I don't know why exactly. 
But removing all `vtimezone` occurrences fixes it (for me).

By the way, thanks for your very useful script.

Manuel

derekantrican:
Thanks for reporting this. I'll look into it

derekantrican:
Obviously `toJSDate()` is a very important method and timezones are very important to maintain. One thing I see that we might be able to improve is rather than using `new ICAL.Timezone(vtimezone);` multiple places, we could probably pass around a global `var timezone = new ICAL.Timezone(vtimezone);` object rather than the `var vtimezone` string. That might lead to a time improvement

boillodmanuel:
During my tests, time was great without using timezone, and Google server
was something like UTC-11 and myself UTC+1.

I still was using toJSDate, but without timezone this method doesn't call
the underlying toUnixTime() method which is very slow (I don't know why).

Le dim. 17 févr. 2019 à 19:30, Derek Antrican <notifications@github.com> a
écrit :

boillodmanuel:
Ok, I spent more time in analysis.

So, as you said, you must register timezone.

Registering once at the beginning is enough :
```js
ICAL.TimezoneService.register(vtimezone);
```

Then, you don't need to set date zone, so you can comment the following lines :
```js
if (vtimezone != null)
  dtstart.zone = new ICAL.Timezone(vtimezone);
...
if (vtimezone != null)
    dtend.zone = new ICAL.Timezone(vtimezone);
```

To fix the very long time, I found a workaround. In cal.js, line 4357, add the following condition to do the work only one time :
```js
if (this.expandedUntilYear >= aYear) {
        return
      }
```
Which will looks like :
```js
    _ensureCoverage: function(aYear) {
      if (this.expandedUntilYear >= aYear) {
        return
      }
      if (ICAL.Timezone._minimumExpansionYear == -1) {
        var today = ICAL.Time.now();
        ICAL.Timezone._minimumExpansionYear = today.year;
      }
```
Each calls was lasting about 5s - after the workaround, only the first call lasts 5s.

The script lasts now 10s against 120s (depending on the number of events)

derekantrican:
[...]

PR #44, opened 2019-05-19
Update Code.gs

jonas0b1011001:
Using Calendar API v3, has to be activated manually.
Added support for Recurrency #7
Added Mail on Eventupdate + Eventdetails in Mailtext #10, #38

Fixed timezone issues -hopefully, working good in all my testing- #29, #34
Fixed All Day Event Handling #40

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -19,11 +19,11 @@
 var targetCalendarName = "";           // The name of the Google Calendar you want to add events to
 var sourceCalendarURL = "";            // The ics/ical url that you want to get events from
 
-var howFrequent = 15;                  // What interval (minutes) to run this script on to check for new events
+var howFrequent = 30;                  // What interval (minutes) to run this script on to check for new events
 var addEventsToCalendar = true;        // If you turn this to "false", you can check the log (View > Logs) to make sure your events are being read correctly before turning this on
 var modifyExistingEvents = true;       // If you turn this to "false", any event in the feed that was modified after being added to the calendar will not update
 var removeEventsFromCalendar = true;   // If you turn this to "true", any event in the calendar not found in the feed will be removed.
-var addAlerts = true;                  // Whether to add the ics/ical alerts as notifications on the Google Calendar events
+var addAlerts = false;                  // Whether to add the ics/ical alerts as notifications on the Google Calendar events
 var addOrganizerToTitle = false;       // Whether to prefix the event name with the event organiser for further clarity 
 var descriptionAsTitles = false;       // Whether to use the ics/ical descriptions as titles (true) or to use the normal titles as titles (false)
 var defaultDuration = 60;              // Default duration (in minutes) in case the event is missing an end specification in the ICS/ICAL file
@@ -69,7 +69,6 @@ var email = "";                        // OPTIONAL: If "emailWhenAdded" is set t
 *
 */
 
-
 //=====================================================================================================
 //!!!!!!!!!!!!!!!! DO NOT EDIT BELOW HERE UNLESS YOU REALLY KNOW WHAT YOU'RE DOING !!!!!!!!!!!!!!!!!!!!
 //=====================================================================================================
@@ -81,7 +80,7 @@ function Install(){
   var validFrequencies = [1, 5, 10, 15, 30];
   if(validFrequencies.indexOf(howFrequent) == -1)
     throw "[ERROR] Invalid value for \"howFrequent\". Must be either 1, 5, 10, 15, or 30";
-
+  
   ScriptApp.newTrigger("main").timeBased().everyMinutes(howFrequent).create();
 }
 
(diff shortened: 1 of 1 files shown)


Code.gs
```diff
@@ -19,11 +19,11 @@
 var targetCalendarName = "";           // The name of the Google Calendar you want to add events to
 var sourceCalendarURL = "";            // The ics/ical url that you want to get events from
 
-var howFrequent = 15;                  // What interval (minutes) to run this script on to check for new events
+var howFrequent = 30;                  // What interval (minutes) to run this script on to check for new events
```
derekantrican: Please don't change default values (such as `howFrequent` or `addAlerts` unless that is the specific issue)

Code.gs
```diff
@@ -310,50 +305,46 @@ function ParseOrganizerName(veventString){
...
-function sameEvent(x){
-  return x.id == this;
-}
-
 function CheckForUpdate(){
   var alreadyAlerted = PropertiesService.getScriptProperties().getProperty("alertedForNewVersion");
   if (alreadyAlerted == null){
     try{
-      var thisVersion = 3.0;
+      var thisVersion = 2.0;
```
derekantrican: I'm assuming this is marked as a change because your branch missed my latest commit https://github.com/derekantrican/GAS-ICS-Sync/commit/e8ab2e363de4d5d1d16295e177ab8727d2da5e8c

Code.gs
```diff
@@ -99,196 +98,192 @@ function DeleteAllTriggers(){
...
     for each (var valarm in valarms){
       var trigger = valarm.getFirstPropertyValue('trigger').toString();
-      event.reminderTimes[event.reminderTimes.length++] = ParseNotificationTime(trigger);
+      event.reminders.overrides[event.reminders.overrides.length++] = ParseNotificationTime(trigger);
     }
   }
-  
+  //RFC allows multiple Rules, Google supports that as well even though it's not displayed in the event details page
+  var recurrenceRules = vevent.getAllProperties('rrule');
+  event.recurrence = [];
```
derekantrican: `recurrence` is never used elsewhere in the code. How does this add recurrence support?

derekantrican:
Please note that I have a large pending pull request (mostly for recurrence support - but also does some large refactoring) here: https://github.com/derekantrican/GAS-ICS-Sync/pull/31 You can go in and review that if you like, but since it has been pending for a while, I was going to do some tests myself soon (when I get some free time). Just haven't gotten around to it.

Otherwise, I think your changes for the timezone fixes and mail additions should be valid! Once my PR is in, this PR will likely have to be condensed down to just those changes.

Please let me know if you have any questions or concerns.

Code.gs
jonas0b1011001: Rules are added on line 285

Code.gs
```diff
@@ -99,196 +98,192 @@ function DeleteAllTriggers(){
...
   }
-
+  
   //------------------------ Add events to calendar ----------------
   if (addEventsToCalendar){
-    Logger.log("Checking " + events.length + " Events for creation")
-    for each (var event in events){
-      if (calendarFids.indexOf(event.id) == -1){
-        var resultEvent;
-        if (event.isAllDay){
```
derekantrican: Why has the `event.isAllDay` check been removed? Am I missing something? Seems like this is a degradation in functionality

Code.gs
derekantrican: Right, the `recurrence` array is populated but nothing ever *uses* the recurrence array in order to create a recurring event

jonas0b1011001: Recurring events are created automatically when inserting an event with rrules (at least that's the case when using Calendar API). No need to create follow on events manually.

Code.gs
jonas0b1011001: The check just moved to line 252. Using the API, all day events are created the same way as 'normal' events so there's no need the check after the new event object has been created.

Code.gs
derekantrican: I'll have to check that. Note that GAS-ICS-Sync does *not* use the full Google Calendar API, but rather uses the [Google Apps Script](https://developers.google.com/apps-script/reference/calendar/) version

jonas0b1011001: I'm aware of that, if changing to the API is not an option for the project my changes most likely are invalid.

derekantrican:
@jonas0b1011001 After reading more of this and seeing what you're doing I see that you're relying on the full Google Calendar API - which is currently not utilized by GAS-ICS-Sync. I've used it myself, so I understand why some things (recurrence, for example) would be easier if done through the full Google Calendar API. But I have avoided using it because it requires extra steps for users (they have to enable it both in the Developers Console and the "Advanced Google Services" window in the script. I want this project to be as easy to use as possible for users so one day we may use the full Google Calendar API, but for now we are sticking with the [Google Apps Script](https://developers.google.com/apps-script/reference/calendar/) version

Issue #23, opened 2019-02-01
Time Zones Off?


rwwest7:
I use the ical's pulled from here:

http://ical.8to18.com/ , specifically this one: http://ical.8to18.com/link/jacksonville/2018-2019/BBB%20V/EDT

I do a manual import to a Google Calendar, and I do a sync with this script to a different calendar. The times for every single event are different by an hour. 

The GAS-ICS-Sync creates the events an hour behind the manual import. Is it possible to tweak the time in the scrypt? They're both using the exact same ICS file.

rwwest77:
I just now shared it. I haven't made any changed below the danger line.

Thank you!

derekantrican:
@rwwest7 I can reproduce this and will investigate

![image](https://user-images.githubusercontent.com/1558019/52213785-ee280c80-2844-11e9-9f9d-7459a383e8f7.png)

derekantrican:
Related to #3

rwwest77:
Can I hard code the timezone in the script? In our case, we don't really need it to dynamically determine. It will always be EST.

derekantrican:
I made a [JSFiddle](http://jsfiddle.net/7f1q56mb/2/) to see if I can reproduce this issue. I can. So I think it is either a problem with ICAL.js parsing the `VTIMEZONE` object or something where I don't know how to use ICAL.js properly

nanobrad:
The VTIMEZONE component is wrong:
```
BEGIN:VTIMEZONE
    TZID:America/New_York
    BEGIN:DAYLIGHT
    DTSTART:19700308T020000
    TZOFFSETFROM:-0400
    TZOFFSETTO:-0500
    RRULE:FREQ=YEARLY;BYDAY=2SU;BYMONTH=3
    TZNAME:EDT
    END:DAYLIGHT
    BEGIN:STANDARD
    DTSTART:19701101T020000
    TZOFFSETFROM:-0500
    TZOFFSETTO:-0400
    RRULE:FREQ=YEARLY;BYDAY=1SU;BYMONTH=11
    TZNAME:EST
    END:STANDARD
END:VTIMEZONE
```
They reversed the TZOFFSET*s. Daylight time in the US for New York (TZOFFSETTO) should be -0400.

Google must be smart enough to not trust the whole VTIMEZONE record, but only pay attention to the TZID.

derekantrican:
[...]

PR #51, opened 2019-08-25
Fixed several timezone issues

### LINKED ISSUE #50: Events show up on wrong timeslot because of timezone
I already added this as a comment to another bug but I assumed it would be more handy to just create a new bug ticket for it.

I'm encountering a similar problem for events as @amytencza in https://github.com/derekantrican/GAS-ICS-Sync/issues/34.

Events that have been created in a different time zone end up at the wrong time slot in the google calendar. In my case the Event is created in India and should apprear at 11:00 (which it does in the original calendar) but in google calendar it appears at 14:30.

When I check the logs I see the following:
The event has a date of **05:30:00 GMT-0700** but India should be at **GMT-5.5**. I tried to look into the code myself but I have the feeling my programming skills are a bit to limited to really be of support here.

I can do some additional debugging if required
Screenshot 2019-08-21 at 09 51 22

jonas0b1011001:
Fix #50, #34, #29

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -98,8 +98,6 @@ function DeleteAllTriggers(){
   }
 }
 
-var vtimezone;
-
 function main(){
   CheckForUpdate();
 
@@ -130,9 +128,9 @@ function main(){
   //Use ICAL.js to parse the data
   var jcalData = ICAL.parse(response);
   var component = new ICAL.Component(jcalData);
-  vtimezone = component.getFirstSubcomponent("vtimezone");
-  if (vtimezone != null)
-    ICAL.TimezoneService.register(vtimezone);
+  var vtimezones = component.getAllSubcomponents("vtimezone");
+  for each (var tz in vtimezones)
+    ICAL.TimezoneService.register(tz);
   
   //Map the vevents into custom event objects
   var events = component.getAllSubcomponents("vevent").map(ConvertToCustomEvent);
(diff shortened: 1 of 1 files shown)


derekantrican:
@jonas0b1011001 I assume you've tested this with a variety of icals? If so, looks good to me and I'll approve it (and probably even make a new update)

jonas0b1011001:
I did, all good except this [one](http://ical.8to18.com/link/jacksonville/2018-2019/BBB%20V/EDT) from #23..still not sure were these events are actually supposed to be..

But feel free to test yourself, i'm sure i have not tested all possibilities.

Code.gs
```diff
@@ -225,7 +223,11 @@ function main(){
 
           if(e.getStartTime().getTime() != fe.startTime.getTime() ||
              e.getEndTime().getTime() != fe.endTime.getTime())
-            e.setTime(fe.startTime, fe.endTime)
+            if (fe.isAllDay){
```
derekantrican: I assume all the 

```
  if (fe.isAllDay){
    e.setAllDayDates(fe.startTime, fe.endTime);
  }else{
    e.setTime(fe.startTime, fe.endTime);
  }
```

is supposed to be within the parent if statement? I'm not super fluent in JS, but do we need brackets around it then? If not, maybe just add them for clarity (also did some minor corrections to conform with conventions in the rest of the program):

```
if(e.getStartTime().getTime() != fe.startTime.getTime() ||
   e.getEndTime().getTime() != fe.endTime.getTime()){
  if (fe.isAllDay)
    e.setAllDayDates(fe.startTime, fe.endTime);
  else
    e.setTime(fe.startTime, fe.endTime);
}
```

jonas0b1011001: > is supposed to be within the parent if statement?
Correct.

Pretty sure brackets are not needed but adding them for clarity is always a good idea

derekantrican: Right. I'm writing up a "coding conventions" document for the wiki right now. Go ahead and push another commit with the suggested change and I'll approve it

PR #52, opened 2019-09-01
Fixed Floating Time Handling

jonas0b1011001:
As mentioned in #34 floating Events should be placed in the target calendar's timezone.


Code.gs
```diff
@@ -267,12 +267,32 @@ function ConvertToCustomEvent(vevent){
...
     
-  event.startTime = icalEvent.startDate.toJSDate();
-  
-  if (icalEvent.endDate == null)
-    event.endTime = new Date(event.startTime.getTime() + defaultDuration * 60 * 1000);
-  else
-    event.endTime = icalEvent.endDate.toJSDate();
+  if (!event.isAllDay && (icalEvent.startDate.tz == null || icalEvent.endDate.tz == null)){
+    Logger.log("Floating Time detected");
+    var targetTZ = CalendarApp.getTimeZone();
```
derekantrican: Should this be `CalendarApp.getTimeZone()` or `targetCalendar.getTimeZone()`? The documentation page has the same description for both (so I'm not sure how they're different) but my inclination tells me it should be the latter

Code.gs
```diff
@@ -267,12 +267,32 @@ function ConvertToCustomEvent(vevent){
...
+    var utcTime = new Date(Utilities.formatDate(jsTime, "Etc/GMT", "HH:mm:ss MM/dd/yyyy"));
+    var tgtTime = new Date(Utilities.formatDate(jsTime, targetTZ, "HH:mm:ss MM/dd/yyyy"));
+    var startTime = new Date(jsTime.getTime() - (tgtTime - utcTime));
+    event.startTime = startTime;
+    jsTime = icalEvent.endDate.toJSDate();
+    utcTime = new Date(Utilities.formatDate(jsTime, "Etc/GMT", "HH:mm:ss MM/dd/yyyy"));
+    tgtTime = new Date(Utilities.formatDate(jsTime, targetTZ, "HH:mm:ss MM/dd/yyyy"));
+    var endTime = new Date(jsTime.getTime() - (tgtTime - utcTime));
+    event.endTime = endTime;
+  }else{
```
derekantrican: Leave the `}` here and put the `else{` on the next line https://github.com/derekantrican/GAS-ICS-Sync/wiki/Code-conventions `#6`

Code.gs
```diff
@@ -267,12 +267,32 @@ function ConvertToCustomEvent(vevent){
...
-  event.startTime = icalEvent.startDate.toJSDate();
-  
-  if (icalEvent.endDate == null)
-    event.endTime = new Date(event.startTime.getTime() + defaultDuration * 60 * 1000);
-  else
-    event.endTime = icalEvent.endDate.toJSDate();
+  if (!event.isAllDay && (icalEvent.startDate.tz == null || icalEvent.endDate.tz == null)){
+    Logger.log("Floating Time detected");
+    var targetTZ = CalendarApp.getTimeZone();
+    Logger.log("Adding Event in " + targetTZ);
```
derekantrican: In general, I think breaking this section up with empty lines and comments would be helpful. Something like:

```
//Get Google calendar timezone
...Code

//Calculate timezone offset
...Code

//Shift floating event times by Google calendar timezone
...Code
```

Code.gs
jonas0b1011001: **CalendarApp.getTimeZone()** gets the default timezone of the overall calendar, i.e. the timezone of the default calendar.
**targetCalendar.getTimeZone()** is the timezone of the target calendar.

These timezones might be the same but don't have to, standard timezone for CalendarApp.createCalendar() is UTC, no matter what your default timezone is.

Initially I used targetCalendar.getTimeZone() as you suggested and seemed logical, the problem is:
Your Calendar is set to PST, target calendar is in UTC and you want to add a event 10am floating. Should it be added to take place at 10am PST or 10am UTC?

When visiting calendar.google.com, the event will be displayed in PST no matter what.
When using CalendarApp.getTimeZone() it will be 10am, using targetCalendar.getTimeZone() it will be displayed at 2am.

derekantrican: So are you talking about a situation like this:

ICS/ICAL calendar: PST *(or is this "undefined"?)*
Google Default Calendar: PST
Target Google Calendar: UTC

It sounds to me like we should still use the target calendar's timezone. If I'm in PST but I have my target calendar in UTC, I should expect a 10am event to show up at 2am. This sounds like what a typical ICS/ICAL syncing situation would do (if the ICS/ICAL was in UTC). 

But, like you mentioned, we should probably also change line 119 `targetCalendar = CalendarApp.createCalendar(targetCalendarName);` to be initialized with the ICS/ICAL timezone (if there is one) using https://developers.google.com/apps-script/reference/calendar/calendar-app#createCalendar(String,Object)

jonas0b1011001: I don't think there is a timezone at vcalendar level. In my fork I set targetCalendar's timezone to the default timezone when creating a new calendar. This kind of avoids the problem as both timezones will be the same and it does not matter which one we choose later on.

Honestly I'm not sure which one is more intuitive to use so I'm fine with changing it to targetCalendar.getTimeZone(). I also don't understand why every calendar has it's own timezone as even if i manually create a Event in targetCalendar and do not specify a timezone, it defaults back to the default calendar timezone and not the target calendar's timezone.

Issue #57, opened 2019-10-05
Execution failed: Event start date must be before event end date.


Zangoran:
Link to my ICS feed (from Canvas, it is a school tool for tracking assignments and stuff):
https://uwplt.instructure.com/feeds/calendars/user_ltI6MNY6ML1ZsYNEL9zwrRBiV38kZcB97NTZagce.ics

This is the error: Execution failed: Event start date must be before event end date. (line 172, file "Code")

I looked through the past issues that looked like this, but couldn't really figure it out. I have the default duration as 60 minutes, not really sure why this doesn't just override the time error.

jonas0b1011001:
Please try [this](https://github.com/derekantrican/GAS-ICS-Sync/issues/56#issuecomment-537569326)!

Zangoran:
@jonas0b1011001 That worked! Thanks!
Is there a reason that isn't merged into the current version yet?

derekantrican:
@jonas0b1011001 *Technically speaking* this format of ics doesn't conform to the [standard](https://tools.ietf.org/html/rfc5545). But we've seen a lot of people have this problem recently so we'll implement the fix for everyone soon.

jonas0b1011001:
In case you want to report the issues to whoever is responsible for creating these ics files (and for the sake of completeness):
in contrast to the example used in [this comment](https://github.com/derekantrican/GAS-ICS-Sync/issues/55#issuecomment-534003159) your events do have the Value=Date property but the date itself is specified in Date-Time format. This (in addition to point 2 in the mentioned comment) is invalid because according to RFC 5545 the

> Value MUST match value type

Zangoran:
Very interesting, yeah I dont think contacting the people with that ics will work.

Issue #56, opened 2019-10-01
All day events (start time = end time) issue


AlbertoM-BostonCollege:
Yahoo calendar creates all day events by adding the same start time and end time to an event. This causes errors on Google: Event start date must be before event end date. at main(Code:172)

The event in question looks like this on the logs:
Title: Apple picking
Id: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
Description:
Start: Sat Oct 05 2019 17:00:00 GMT-0700   (PDT)
End: Sat Oct 05 2019 17:00:00 GMT-0700   (PDT)

jonas0b1011001:
Please take a look at #55 and try the workaround suggested there.

AlbertoM-BostonCollege:
error doesn't seem to be with parsing dates, so it doesn't adjust end date., so I assumed an all day event is one that starts and ends 24hrs later.  I added this but didn't quite cut it. 

```
 //checking if Startdate equals Enddate
  if (icalEvent.startDate == icalEvent.endDate)
    Logger.log("Start Date and End Date are the same. Adjusting End Date + 1");
    icalEvent.endDate.adjust(1,0,0,0)
```

jonas0b1011001:
Try replacing your code.gs with the following [code](https://pastebin.com/raw/ksTUnGJD) and try again :)

AlbertoM-BostonCollege:
Yup, that did it!
Thanks for the quick reply!

derekantrican:
@jonas0b1011001 Sounds like this might become enough of a problem that we should just account for this in the main code. Something like:

```
if (duration is 0){
  if (start is date)
    //change duration to 1 day
  else //start is datetime
    //change duration to the "defaultDuration" setting
}
```

What do you think?

jonas0b1011001:
Definitely seems to be a more common error than i thought it would be and might be worth pushing to the repo.

Issue #55, opened 2019-09-22
When VEVENT is a "whole day" event the script ends in error


BompaBeer:
Fout
22 sep. 2019 20:59:53
Error: invalid date-time value: "2019-10-04T::"
    at [unknown function](ical.js:5834)
    at [unknown function](ical.js:760)
    at [unknown function](ical.js:2888)
    at [unknown function](ical.js:2871)
    at [unknown function](ical.js:3017)
    at [unknown function](ical.js:2449)
    at ConvertToCustomEvent(Code:264)
    at main(Code:138)

jonas0b1011001:
Seems like your source calendar is not formatted properly. Could you share the event in question?
If it is a all-day event time definition should look like "DTSTART;VALUE=DATE:YYYYMMDD"

BompaBeer:
Thank you for your quick answer.
here is an example of an all-day event the way I find it in the iCal file

BEGIN:VEVENT
DTSTART:20191004
DTEND:20191004
UID:gimme-PeMomhZrB2g3w63bY
SUMMARY:Saved by the bell
DESCRIPTION:Goed onderwijs voor iedereen! Link:https://app.gimme.eu/mygimme
END:VEVENT

greetings from Belgium

jonas0b1011001:
So there are actually two issues with the source.
1. Start and End Date miss the Value=Date property.
As
> The default value type is DATE-TIME.

the script is expecting a value that includes the time.

2. Start and End Date are the same day.

> [...]its (DTEND) value MUST be later in time than the value of the "DTSTART"
      property.

> The "DTEND" property for a "VEVENT" calendar component specifies the non-inclusive end of the event.

After correcting these error, the valid event would look like this:
`BEGIN:VEVENT`
`DTSTART;VALUE=DATE:20191004`
`DTEND;VALUE=DATE:20191005`
`UID:gimme-PeMomhZrB2g3w63bY`
`SUMMARY:Saved by the bell`
`DESCRIPTION:Goed onderwijs voor iedereen! Link:https://app.gimme.eu/mygimme`
`END:VEVENT`

DTEND can be omitted if the event takes place at a single  day though.
> For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE value type but no "DTEND" nor "DURATION" property, the event's duration is taken to be one day.

Therefore the following would also be valid:
`BEGIN:VEVENT`
`DTSTART;VALUE=DATE:20191004`
[...]

PR #58, opened 2019-10-06
Workaround for invalid allday event definition (dtstart = dtend)

jonas0b1011001:
Fix for #55 #56 #57

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -26,7 +26,6 @@ var removeEventsFromCalendar = true;   // If you turn this to "true", any event 
 var addAlerts = true;                  // Whether to add the ics/ical alerts as notifications on the Google Calendar events
 var addOrganizerToTitle = false;       // Whether to prefix the event name with the event organiser for further clarity 
 var descriptionAsTitles = false;       // Whether to use the ics/ical descriptions as titles (true) or to use the normal titles as titles (false)
-var defaultDuration = 60;              // Default duration (in minutes) in case the event is missing an end specification in the ICS/ICAL file
 
 var emailWhenAdded = false;            // Will email you when an event is added to your calendar
 var email = "";                        // OPTIONAL: If "emailWhenAdded" is set to true, you will need to provide your email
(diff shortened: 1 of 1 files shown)


Code.gs
```diff
@@ -270,6 +269,11 @@ function ConvertToCustomEvent(vevent){
   if (icalEvent.startDate.isDate && icalEvent.endDate.isDate)
     event.isAllDay = true;
   
+  if (icalEvent.startDate.compare(icalEvent.endDate) == 0 && event.isAllDay){
+    //Adjust dtend in case dtstart equals dtend as this is not valid for allday events
+    icalEvent.endDate = icalEvent.endDate.adjust(1,0,0,0);
+  }
```
jonas0b1011001: This is only required for all day events as 

> For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE-TIME value type but no "DTEND" property, the event ends on the same calendar date and time of day specified by the "DTSTART" property.

Code.gs
```diff
@@ -292,12 +296,7 @@ function ConvertToCustomEvent(vevent){
   }
   else{
     event.startTime = icalEvent.startDate.toJSDate();
-    if (icalEvent.endDate == null){
```
jonas0b1011001: Obsolete as icalEvent.endDate will never be null. If the vevent is missing dtend ical.js will calculate an appropriate dtend according duration property (if it exists).

If duration is not specified
- dtend will be dtstart if Value=Date-Time
- dtend will be dtstart + 1d if Value=Date

Code.gs
```diff
@@ -26,7 +26,6 @@ var removeEventsFromCalendar = true;   // If you turn this to "true", any event
 var addAlerts = true;                  // Whether to add the ics/ical alerts as notifications on the Google Calendar events
 var addOrganizerToTitle = false;       // Whether to prefix the event name with the event organiser for further clarity 
 var descriptionAsTitles = false;       // Whether to use the ics/ical descriptions as titles (true) or to use the normal titles as titles (false)
-var defaultDuration = 60;              // Default duration (in minutes) in case the event is missing an end specification in the ICS/ICAL file
```
jonas0b1011001: There is no real need for a default Duration as Google Calendar accepts Events with 0min Duration (which is perfectly fine according RFC as well - see other comments).

derekantrican: My original thought with `defaultDuration` was that the user could specify a duration in the case of "non-all-day events" where the duration and end datetime are not specified. Do you think we should just allow events to be 0 minutes in that case?

jonas0b1011001: I don't really see the need of defining a own default while the RFC already has one.
I would totally agree if ical did not allow to define a custom duration for events, but as of now i think we should aim for creating an as accurate representation of the input as possible, which includes 0 minute events.

derekantrican:
@jonas0b1011001 I'm assuming you've put these changes into your repo as well?

jonas0b1011001:
Good point, just did

PR #60, opened 2019-10-07
Several new features

rufuswilson:
Multi Target Calendar
Support for multiple target calendar has been added

Multi Source Calendar per Target Calendar
Each target calendar can be the union of several source URLs

Ignore Titles
It is possible to ignore events based on regexes applied onto the titles

Timeout
It graciously handles timeout limit (6 minutes) by limiting each run to a maximum of 275s and running the loop function every 300s
(One of my calendars has 670 events and I can only process 300 in 275s)

Quota error
Every time a quota error is reached, the loop is stoped and waits for the next full update.

Only future events
Adds option to only sync events that have a start date after the beginning of the sync process

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -122,7 +122,6 @@ function Reset(){
 
 
 function Loop(){
-
   try{
     //------------------------ Error checking ------------------------
     if (emailWhenAdded && email == "")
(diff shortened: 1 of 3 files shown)


derekantrican:
Thanks for all the suggestions!

Unfortunately, some of these features are already taken care of. I think the next release will be upgrading the script to the full Google Calendar API (for a few reasons, but mostly to easily support recurring events) and that's being managed in @jonas0b1011001 's fork here: https://github.com/jonas0b1011001/GAS-ICS-Sync-CalAPIv3

His fork covers a couple of the features you are suggesting including Multiple Calendar support, handling quota errors (through exponential backoff - which is the better approach here), and I believe it also fixes most of the timeout issues. The "ignore titles based on regex" and "only future events" features seem interesting, though. And code refactoring is likely very helpful!

Please take a look at his fork and see how your changes might fit in. You can also discuss these features with both of us via the [discord](https://discord.gg/DRBpb4k).

derekantrican:
Thanks for all the suggestions!

Unfortunately, some of these features are already taken care of. I think the next release will be upgrading the script to the full Google Calendar API (for a few reasons, but mostly to easily support recurring events) and that's being managed in @jonas0b1011001 's fork here: https://github.com/jonas0b1011001/GAS-ICS-Sync-CalAPIv3

His fork covers a couple of the features you are suggesting including Multiple Calendar support, handling quota errors (through exponential backoff - which is the better approach here), and I believe it also fixes most of the timeout issues. The "ignore titles based on regex" and "only future events" features seem interesting, though. And code refactoring is likely very helpful!

Please take a look at his fork and see how your changes might fit in. You can also discuss these features with both of us via the [discord](https://discord.gg/DRBpb4k).

PR #61, opened 2019-10-08
Add "removeExistingEvents" flag

huang-a:
The flag is by default false. If the flag is set to true, "removeEventsFromCalendar" will also apply to events that existed before the script is installed or added via the Google Calendar interface(s) after the script is installed.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -23,6 +23,7 @@ var howFrequent = 15;                  // What interval (minutes) to run this sc
 var addEventsToCalendar = true;        // If you turn this to "false", you can check the log (View > Logs) to make sure your events are being read correctly before turning this on
 var modifyExistingEvents = true;       // If you turn this to "false", any event in the feed that was modified after being added to the calendar will not update
 var removeEventsFromCalendar = true;   // If you turn this to "true", any event in the calendar not found in the feed will be removed.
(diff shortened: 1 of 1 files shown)


derekantrican:
Unfortunately, I think this will could create more problems than it solves. I think there will be too many people getting confused as to what this means and then getting frustrated when they can't create events on that calendar via the Google Calendar web interface (or worse: wiping out their entire Google Calendar).

So I think this should just be something you implement own your own version.

huang-a:
That's reasonable. My use case is to mirror the ical feed onto Google Calendar to utilize its UI. Since I'm using this script to mirror just the feed and nothing else, I was surprised when events that I added manually (accidentally to the wrong calendar), stayed there after the script ran.

Is there a license you want to put this codebase under? I'm not sure what licenses I can use when I open source my own version.

derekantrican:
Yeah, we should probably add a license. I'm not super knowledgeable in that area - I think the GNU GPLv3 license looks pretty good. Do you have any suggestions?

huang-a:
It really depends on how you want your code to be distributed. I use (this site)[https://choosealicense.com/] to choose licenses. GPLv3 is a good license.

derekantrican:
Ok, I think we'll go with that one. I'm going to go ahead to close this PR because we've decided not to include the changes

Issue #62, opened 2019-10-15
Eventually, should support mapping between multiple ICS calendars and Google Calendars


derekantrican:
Some day in the future a user should have the ability to add the urls for as many ICS calendars as they want and choose the Google calendar(s) that they sync to. Something like:

test.com/cal1.ics --> Personal
test.com/cal2.ics --> Work, Travel

After we make our initial "multiple calendars support" changes this will somewhat be possible, but only if you're running multiple instances of the program (as we only support 1 ics per instance). Eventually this should be possible with a single instance.

derekantrican:
### Closing as this is supported by the latest release: [v5.0](https://github.com/derekantrican/GAS-ICS-Sync/releases/tag/5.0)

*If you find that there are still bugs, please comment on this issue and let us know*

PR #63, opened 2019-10-16
Latest 'stable' version

jonas0b1011001:
- Multicalendar support, merge multiple ical calendars to various google calendars
      Option to add sourcecalendar name to event title (if available)
- Support of VTODO-Elements
- Full Recurrence support
      RRULE, EXRULE, RDATE, EXDATE, RECURRENCE-ID
- Full mapping of ICAL properties to Google Calendar Event Properties
      Attendees (Name, Mail, Response), Status, Sequence, Class, Transp, URL
- EMail notification when an event is updated by the script
- Exponential Backoff to handle API-throttling
- Support any Interval as trigger

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -22,10 +22,12 @@ function Install(){
   
   //Custom error for restriction here: https://developers.google.com/apps-script/reference/script/clock-trigger-builder#everyMinutes(Integer)
   var validFrequencies = [1, 5, 10, 15, 30];
-  if(validFrequencies.indexOf(howFrequent) == -1)
-    throw "[ERROR] Invalid value for \"howFrequent\". Must be either 1, 5, 10, 15, or 30";
-  
-  ScriptApp.newTrigger("startSync").timeBased().everyMinutes(howFrequent).create();
+  if (howFrequent < 1)){
+    throw "[ERROR] \"howFrequent\" must be greater than 0.";
+  }
+  else{
+    ScriptApp.newTrigger("startSync").timeBased().after(howFrequent * 60 * 1000).create();
+  }
 }
 
 function Uninstall(){
--- a/Code.gs
+++ b/Code.gs
@@ -214,7 +214,13 @@ function main(){
         }
       }
       if (event.hasProperty('status')){
-        newEvent.status = event.getFirstPropertyValue('status').toString().toLowerCase();
+        var status = event.getFirstPropertyValue('status').toString().toLowerCase();
+        if (["confirmed", "tentative", "cancelled"].indexOf(status) > -1)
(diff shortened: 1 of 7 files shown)


derekantrican:
Couple notes for when we go to merge this into master:

- We'll need to undo your README changes
- Also, I'll need to create a new script file for the app to be hosted from so that it is easier for users to auth the full API (this is easier for scripts created [after April 8, 2019](https://developers.google.com/apps-script/guides/services/advanced#requirements) which the current GAS-ICS-Sync file was not)
- I think the above point will also simplify the install instructions so those will have to be updated
- While I'm looking at the big block comment at the top of the file I noticed you have the settings in a separate settings file. I think this is great for making an HTML UI, but I think while we're having people edit the code to set settings, maybe we just keep the whole "instructions/settings/info" section at the top of Code.gs the same as it currently is.

jonas0b1011001:
> We'll need to undo your README changes

Do you know if there is a way to exclude it from the merge?

> Also, I'll need to create a new script file for the app to be hosted from so that it is easier for users to auth the full API (this is easier for scripts created after April 8, 2019 which the current GAS-ICS-Sync file was not)

Could you 'install' the script linked in the readme of this branch to see which steps are actually needed to get this version to run?

> While I'm looking at the big block comment at the top of the file I noticed you have the settings in a separate settings file. I think this is great for making an HTML UI, but I think while we're having people edit the code to set settings, maybe we just keep the whole "instructions/settings/info" section at the top of Code.gs the same as it currently is.

Yeah I got rid of the descriptive section at some point, will have to readd the whole top section.

derekantrican:
I don't think there's any way to really exclude a change from a PR. I think the changes will just have to be undone in the source branch.

Your README's linked script seems to install great! In fact, you don't need step 3 of your install instructions (that's the real benefit we get now that Google changed this process)

jonas0b1011001:
I restored the readme/header section and edited the Instructions.

I think this includes #62 ?

derekantrican:
I think so? Since your script is using the full API I'll admit that I'm a little lost trying to understand it (I'm not used to the full API through JavaScript, just through C#). My main question: I know there is a possible problem with the current script of syncing multiple ICS files to the same target calendar (because one ICS file would think the other ICS' events are old and delete them) - does your script version have a similar issue? Or is it fine?

*Note that there may be a conflict with #65 that I just noticed you also have a similar change for ("asynchronously" starting  the `main` function on install)*

jonas0b1011001:
There should be no issues. I'm collecting all events from the sources before starting to process them, so there's actually no difference if you sync a single or multiple calendars as it will always be a single array with all events.

PR #65, opened 2019-10-17
Small improvements

bestknighter:
Hey! I've noticed a room for two small improvements at your already wonderful script (that was exactly what I was looking for) and decided to do it myself. I don't even know if this is PR-worthy but here it is.

**- Removes a possible failure point**
This one is more useful to guarantee that it may work in the future without any more tweaks. So it's great for other people, in the future. For us is quite useless. But who knows? I won't allow the millennium bug to happen again. :laughing:

**- First run at installation**
When I installed the script, I expected to have the events show up in my calendar right away. I was a bit confused to why that wasn't happening. A quick look at the code made me notice that the first run would be at "howFrequent" minutes after installing. That is not intuitive but works fine.
I just added a fire-once trigger to execute 1s after installing so the results may show "immediately" without making the installation slower. Because of how triggers work, it may run between after installation and 1 minute later. People who don't know how to code will be able to see the results a few moments after installing instead of thinking that the script didn't work. Of course, the size of the calendar will affect how quickly they are going to show up but that's beyond our control.

If you don't want to have a disabled trigger in the project, it's quite easy to make a routine to clean that up. I can do that if you want but I don't think it's necessary.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -158,7 +158,9 @@ function main(){
   //----------------------------------------------------------------
 
   if(addEventsToCalendar || removeEventsFromCalendar){
-    var calendarEvents = targetCalendar.getEvents(new Date(2000,01,01), new Date( 2100,01,01 ))
+    var currentYear = new Date().getFullYear();
+    // If someone does care about an event from way before they were even born or that will happen way after their death, then just increase these ranges
+    var calendarEvents = targetCalendar.getEvents(new Date(currentYear-100,01,01), new Date(currentYear+100,01,01 ));
     var calendarFids = []
     for (var i = 0; i < calendarEvents.length; i++)
       calendarFids[i] = calendarEvents[i].getTag("FID");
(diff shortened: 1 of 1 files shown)


Code.gs
```diff
@@ -83,7 +83,8 @@ function Install(){
   if(validFrequencies.indexOf(howFrequent) == -1)
     throw "[ERROR] Invalid value for \"howFrequent\". Must be either 1, 5, 10, 15, or 30";
 
-  ScriptApp.newTrigger("main").timeBased().everyMinutes(howFrequent).create();
+  ScriptApp.newTrigger("main").timeBased().after(1000).create(); // Runs the script right after installing
```
derekantrican: Why do we need a trigger for this? Is there a problem with just calling `main();`?

bestknighter: It's more about being user friendly with those who don't understand how Script Apps, or coding in general, works.

If the calendar has lots of events, the installation itself may take too long. Calling `main();` does the same thing as with trigger, but the "Running function install..." will not disappear unless it has finished syncing.
With triggers, it disappears quickly and the function main will be "running in the background".

We know that, effectively speaking, it's the same but the latter allows the user to close the tab without worries that it will interrupt the installation process, even though it wouldn't interrupt anyway (as far as I know).

derekantrican: Ah, right. It's a way to do "async" in Google Apps Script. Good point

PR #78, opened 2019-11-05
Remove triggers from previous versions

jonas0b1011001:
As mentioned in Discord support channel


derekantrican:
@jonas0b1011001 Before you put this in, could you also resolve this as well (should be a simple change): https://github.com/derekantrican/GAS-ICS-Sync/issues/94#issuecomment-576461428

I think it has to do with some specific cases with syncing Outlook calendars

jonas0b1011001:
This PR should now include all fixes/changes from bugreports since we launched the new version. Might be a good time to push an update as there are quite a few changes.

derekantrican:
Yeah, agreed

derekantrican:
Go ahead and update the version number at the bottom of Helpers.gs to whatever you think is appropriate (whether major or minor version)

PR #99, opened 2020-02-10
Fix for duplicate events randomly occurring

Ali1:
Somehow over the weeks, it seems once in a while, a sync event just creates duplicates of every single event in the ics file. These will never be deleted by the script.

Now sure if anyone has also experienced this but it's difficult to reproduce. I'm thinking it's some weird calendar read error when the script thinks there are no events in the target calendar at a given moment.

This PR introduces a more aggressive clean up which also deletes any duplicate events in the target calendar.

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -507,21 +507,25 @@ function processEventInstance(recEvent){
 
 /**
  * Deletes all events from the target calendar that no longer exist in the source calendars.
+ * Also deletes duplicate events on target calendar that may have been introduced by a script read error
  * If onlyFutureEvents is set to true, events that have taken place since the last sync are also removed.
  */
 function processEventCleanup(){
+  var processedIndex = []; // ensure no duplicates on Google Calendar
   for (var i = 0; i < calendarEvents.length; i++){
       var currentID = calendarEventsIds[i];
       var feedIndex = icsEventsIds.indexOf(currentID);
       
-      if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
-        Logger.log("Deleting old event " + currentID);
+      if(processedIndex.indexOf(currentID) != -1 || feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
+        Logger.log("Deleting event " + currentID + " (clean up)");
         try{
           Calendar.Events.remove(targetCalendarId, calendarEvents[i].id);
         }
         catch (err){
           Logger.log(err);
         }
+      } else {
+        processedIndex.push(currentID);
       }
     }
 }
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -516,7 +516,7 @@ function processEventCleanup(){
       var currentID = calendarEventsIds[i];
       var feedIndex = icsEventsIds.indexOf(currentID);
       
-      if(processedIndex.indexOf(currentID) != -1 || feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
+      if((feedIndex  == -1 && calendarEvents[i].recurringEventId == null) || processedIndex.indexOf(currentID) != -1){
         Logger.log("Deleting event " + currentID + " (clean up)");
         try{
           Calendar.Events.remove(targetCalendarId, calendarEvents[i].id);


Helpers.gs
```diff
@@ -507,21 +507,25 @@ function processEventInstance(recEvent){
...
  */
 function processEventCleanup(){
+  var processedIndex = []; // ensure no duplicates on Google Calendar
   for (var i = 0; i < calendarEvents.length; i++){
       var currentID = calendarEventsIds[i];
       var feedIndex = icsEventsIds.indexOf(currentID);
       
-      if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
-        Logger.log("Deleting old event " + currentID);
+      if(processedIndex.indexOf(currentID) != -1 || feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
```
derekantrican: Can you make the boolean operator precedence clear here? (Add parentheses) I think the intention is `(feedIndex  == -1 && calendarEvents[i].recurringEventId == null) || processedIndex.indexOf(currentID) != -1`, correct?

derekantrican:
You said "somehow". Do you know why these duplicate events get created and why they don't get deleted on the next iteration? We'll want to make sure that this isn't an issue specific to your copy of the code or ICS and is applicable to everyone

Ali1:
Nope, I had not made any changes to the code and I can't pin down what happened at the time they got introduced. All I know is over the last few weeks, this issue occurred twice on all the calendars in the sourceCalendars therefore resulting in 3 copies of every single event remaining on the calendar.

Why would they then get deleted? The next iteration does not have code that deletes duplicate entries. With this PR, they would

Helpers.gs
Ali1: sure

derekantrican:
Do you mind either sharing your ICS here or sending it to me at derekantrican@gmail.com?

Ali1:
Can't sorry I'm in healthcare.

Ali1:
You can leave it open to see if anyone else experiences issues with duplicate events if you want.

Or you can merge and if this is not a real issue, then it will do nothing.

But I think users of the script do expect that when they put true for removeEventsFromCalendar, they expect that any duplicates get removed too no matter how or why they got introduced.

By the way, I just remembered I have a URL entry to the script which gets triggered when my ics gets updated - so perhaps two versions of the script were running at the same time - this may be the cause for the duplicates to get introduced.

derekantrican:
>By the way, I just remembered I have a URL entry to the script which gets triggered when my ics gets updated - so perhaps two versions of the script were running at the same time - this may be the cause for the duplicates to get introduced.

Yeah, that does sound possible. Something else I was thinking of: we've had a few different users where their ICS won't keep the same event UID from iteration to iteration. You can test this yourself by downloading your ICS once, then downloading it again and seeing if there are any of the same UIDs between those copies. If not, then that is another reason why there might be duplicates. (Let me know if that doesn't make sense)

Ali1:
No my id's are intentionally set and fixed

jonas0b1011001:
How do we make sure that the latest version of an event stays in the calendar? Keeping an old version might be worse than having duplicates.

Ali1:
Both versions of the duplicate should be correct I think...

But to answer your question, by setting the right sort order for retrieving calendar events from target calendar (latest addition first), this will ensure the most up to date one remains

kghost:
I have experienced duplicated event, by running the script multiple times simultaneously.

When I importing a very large ical, the add event may last a few minutes, during the process, if you start another run, there will be duplicated event, and they are not going to be removed forever.

jonas0b1011001:
> I have experienced duplicated event, by running the script multiple times simultaneously.

We are aware that simultaneous runs cause duplicates. However the latest version should not allow you to run multiple syncs at the same time. Could you please try if that's enough to solve your problem?

derekantrican:
I think this is stale so I'm going to close it. Please comment here again if that is not the case

Ali1:
I deleted my fork just wanted to leave this in case it gets removed.
I don't think the issue occurs anymore

image
```
 * Also deletes duplicate events on target calendar that may have been introduced by a script read error
 * If onlyFutureEvents is set to true, events that have taken place since the last sync are also removed.
 */
function processEventCleanup(){
  var processedIndex = []; // ensure no duplicates on Google Calendar
  for (var i = 0; i < calendarEvents.length; i++){
      var currentID = calendarEventsIds[i];
      var feedIndex = icsEventsIds.indexOf(currentID);

      if((feedIndex  == -1 && calendarEvents[i].recurringEventId == null) || processedIndex.indexOf(currentID) != -1){
        Logger.log("Deleting event " + currentID + " (clean up)");
        try{
          Calendar.Events.remove(targetCalendarId, calendarEvents[i].id);
        }
        catch (err){
          Logger.log(err);
        }
      } else {
        processedIndex.push(currentID);
      }
    }
}
```

PR #101, opened 2020-02-17
V8 runtime

### LINKED ISSUE #98: Support new V8 runtime
SyntaxError: Unexpected identifier (line 129, file "Code.gs") appears and won't let me install.

Line 129 says:

 for each (var calendar in sourceCalendars){

No idea what that's about.

derekantrican:
Resolves #98


derekantrican:
@jonas0b1011001 Decided to just revert `timeZone` to default (`"GMT"`) since that's what it will get set to if we don't include the `timeZone` section. Also kept `exceptionLogging` section as it may be helpful later

PR #121, opened 2020-04-22
Added DateTime support for reminders

### LINKED ISSUE #120: Support DATE-TIME triggers for notifications
`parseNotificationTime` https://github.com/derekantrican/GAS-ICS-Sync/blob/24fc1448d2afe977025d4c92524010b5c7e92aaa/Helpers.gs#L706 Should support the following scenario:

```
BEGIN:VEVENT
DTSTAMP:20200422T174526Z
DTSTART:20200426T080000Z
DTEND:20200426T090000Z
SUMMARY:asdfasdfasdf
BEGIN:VALARM
TRIGGER;VALUE=DATE-TIME:20200426T073000Z
ACTION:DISPLAY
UID:94f8623f-898a-4f00-96dd-da9692ed0359
DESCRIPTION:asdfasdfasdf
X-APPLE-DEFAULT-ALARM:TRUE
X-WR-ALARMUID:94f8623f-898a-4f00-96dd-da9692ed0359
END:VALARM
END:VEVENT
```

jonas0b1011001:
Fix #120 
Changed parseNotificationTime to ignore seconds and return minutes as GCal does not support seconds anyways.

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -366,8 +366,12 @@ function createEvent(event, calendarTz){
       var overrides = [];
       for (var valarm of valarms){
         var trigger = valarm.getFirstPropertyValue('trigger').toString();
+        try{
+          var alarmTime = new ICAL.Time.fromString(trigger);
+          trigger = alarmTime.subtractDateTz(icalEvent.startDate).toString();
+        }catch(e){}
         if (overrides.length < 5){ //Google supports max 5 reminder-overrides
-          var timer = parseNotificationTime(trigger)/60;
+          var timer = parseNotificationTime(trigger);
           if (0 <= timer && timer <= 40320)
             overrides.push({'method' : 'popup', 'minutes' : timer});
         }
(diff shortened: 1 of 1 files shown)


Helpers.gs
```diff
@@ -701,7 +705,7 @@ function parseAttendeeResp(veventString){
  * Will return 0 by default.
  *
  * @param {string} notificationString - The string to parse
- * @return {number} The notification time in seconds
+ * @return {number} The notification time in minutes
  */
 function parseNotificationTime(notificationString){
```
derekantrican, line 710: What was the rationale for changing this to minutes? Is it not possible for a notification trigger to contain `__S` or did we just decide to not support it since Google Calendar doesn't support a number of seconds for a notification?

jonas0b1011001: I'm sure a trigger could include seconds (even though i have not found a definition of all valid options) but google will not add these event.

Triggers like **-PT30M30S** currently throw an error at _Calendars.insert_, with the change it would just omit the seconds and add a reminder 30min prior to the event, that's as close as we can get..

derekantrican: Sounds like a good assessment.

PR #122, opened 2020-04-22
Execution summary email

### LINKED ISSUE #87: Email of new items converted into "digest"
The email alert for new entries is a good idea, but sometimes creates large numbers of entries if multiple new diary entries added.

How about condensing the appointments into a single list and adding them to the text of one single email e.g.

To: [email address]
Subject: [NN] new appointment(s) added to [CalendarName] 
Contents:-
The following appointments have been added by GAS-ICS Sync:-
[Date] [Title] [Location] [Time]
[Date] [Title] [Location] [Time]
[Date] [Title] [Location] [Time]
[Date] [Title] [Location] [Time]
[Date] [Title] [Location] [Time]
[Date] [Title] [Location] [Time]
etc

jonas0b1011001:
Fix #87

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -38,8 +38,7 @@ var defaultAllDayReminder = -1;        // Default reminder for all day events in
                                        // See https://github.com/derekantrican/GAS-ICS-Sync/issues/75 for why this is neccessary.
 var addTasks = false;
 
-var emailWhenAdded = false;            // Will email you when an event is added to your calendar
-var emailWhenModified = false;         // Will email you when an existing event is updated in your calendar
+var emailSummary = false;              // Will email you when an event is added/modified/removed to your calendar
 var email = "";                        // OPTIONAL: If "emailWhenAdded" is set to true, you will need to provide your email
 
 /*
@@ -114,6 +113,9 @@ var calendarEventsIds = [];
 var icsEventsIds = [];
 var calendarEventsMD5s = [];
 var recurringEvents = [];
+var addedEvents = [];
+var modifiedEvents = [];
+var removedEvents = [];
 var startUpdateTime;
 
 function startSync(){
@@ -130,7 +132,7 @@ function startSync(){
     startUpdateTime = new ICAL.Time.fromJSDate(new Date());
   
   //Disable email notification if no mail adress is provided 
-  emailWhenAdded = emailWhenAdded && email != "";
+  emailSummary = emailSummary && email != "";
   
   sourceCalendars = condenseCalendarMap(sourceCalendars);
   for (var calendar of sourceCalendars){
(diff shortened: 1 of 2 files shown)


yellowbarchetta:
I copied across helpers.gs and code.gs to incorporate this, but had the following issues:-
I needed to define addedEvents, modifiedEvents and removedEvents, I added them in to code.gs just before the function startSync is drafted (not sure if that is "correct" or not, but it worked - var addedEvents = []; etc

I also needed to amend the line in helpers.gs from MailApp.send to GmailApp.send and replace the single variable "message" with (email, subject, "", {htmlBody: body}) as I think MailApp needs api authorisation (or similar) that I don't have?

yellowbarchetta:
Aha, you're right of course - I'd missed those added variables when I'd copied across (I've made other adjustments above that in the code so I couldn't just drop it all in, but overlooked the changes in the variables.  It did seem odd that that error hadn't come up at your end!

Everywhere else I think the code uses / used GmailApp rather than MailApp, but happy to authorise both approaches - GmailApp works fine though, nicely formatted output etc., and as I've added a tiny bit of separate code to give me a daily heartbeat as to whether the service is still running (as I don't trust the Google Apps Script notifications) whcih uses GmailApp it makes sense to me to leave that as the only route.

jonas0b1011001:
Without having spoken with Derek about this, i plan to completely switch to MailApp. #123 is also talking about this, there is simply no valid reason for us to request full access to someone's inbox (even though we do not utilize these permissions).
GMailApp and MailApp should be fully interchangeable with no difference in regards to sending mails .

yellowbarchetta:
Makes sense. I've edited back to MailApp, I think the problem I was seeing was that MailApp causes emails to skip the inbox when sent to / from the same gmail account, and as I wasn't seeing them in the inbox I didn't think to check the sent items.  Now I know I can confirm it's working as designed anyway.  

However, if anyone can suggest a fix to stop them skipping the inbox in the gmail client, that would be helpful as (for me) the purpose of the emails is to highlight changes that might otherwise skip my attention (one of my employer's PAs often makes entries in my calendar, so it's nice to get alerted on my phone via Gmail when that happens).

jonas0b1011001:
That's weird, the mails show up just fine in my inbox.
image

After a [quick search](https://stackoverflow.com/questions/20929322/sending-email-to-myself-through-gas-stays-in-gmail-sent-box-and-not-in-inbox) , this seems to be a rare but known issue..

derekantrican:
Here's a related issue: https://issuetracker.google.com/issues/36763377 (it's Asthough note that it's about 4.5 years old at this point. Google is not exactly the fastest at fixing some of these "user-reported" bugs)

Issue #149, opened 2020-08-21
TypeError: Cannot read property 'dateTime' of undefined


dfishburn:
- I am getting periodic:
`TypeError: Cannot read property 'dateTime' of undefined in Helper.gs` @ line 492 in version 5.3

When reading a recurring calendar event created in CEST, while I am in EST (if relevant).

So I modified the code and changed these this:
```
  var eventInstanceToPatch = calendarEvents.filter(function(item){
      var origStart = item.originalStartTime.dateTime || item.originalStartTime.date;
      var instanceStart = new ICAL.Time.fromString(origStart);
      
      return (instanceStart.compare(recIDStart) == 0);
  });
```
To this:
```
  var eventInstanceToPatch = calendarEvents.filter(function(item){
    try{
      var origStart = item.originalStartTime.dateTime || item.originalStartTime.date;
      var instanceStart = new ICAL.Time.fromString(origStart);
      
      return (instanceStart.compare(recIDStart) == 0);
    }
    catch(error){
      Logger.log("eventInstanceToPatch - Invalid date");
      Logger.log(error); 
      Logger.log("eventInstanceToPatch - item");
      Logger.log(item); 
      Logger.log("eventInstanceToPatch - repeat the failure");
      var origStart = item.originalStartTime.dateTime || item.originalStartTime.date;
    }
  });
```
- I didn't know the syntax to rethrow the error, so I just called the same origStart line.  Now I get to see the error and log the item that is causing it:
```
TypeError: Cannot read property 'dateTime' of undefined
    at [unknown function](Helpers:503:46)
    at processEventInstance(Helpers:492:45)
    at startSync(Code:205:7)
```

The item which triggered the error:
[...]

Issue #135, opened 2020-06-15
TypeError: Cannot read property 'dateTime' of undefined (line 493, file "Helpers")


deanosauras:
- Description of the problem
Outlook to Google calendar using ics link is showing attached error. However, it seems to be working as far as syncing most of the events in the calendar. I don't know if it is just warning or a genuine error that doesn't sync all the events in my calendar. I am using 5.2 script.

Is it possible that this is caused by a recurring calendar event on Outlook. And one for the event dates in the series has a different time?

Not a duplicate of #133 #115 

- Screenshots
![2020-06-15 14_04_01-Outlook to Google Calendar Sync](https://user-images.githubusercontent.com/66968127/84690822-1fbce780-af11-11ea-88c3-3067b4e4803e.jpg)

![image](https://user-images.githubusercontent.com/66968127/84692851-855ea300-af14-11ea-9cea-9b35246f337a.png)

- An input ics string that can reproduce the problem
["https://outlook.office365.com/owa/calendar/6cc9c9ba09864c0b84ff0552e83c8b1a@futurevaultinc.com/ccc41ae7ff424d6796284efdc72250bb12804937182588779247/calendar.ics", "FutureVault Calendar"]

You can see [this issue](https://github.com/derekantrican/GAS-ICS-Sync/issues/34) for an example of good information when submitting a bug

jonas0b1011001:
I don't get an error with 5.3, try updating to the latest version.

deanosauras:
> I don't get an error with 5.3, try updating to the latest version.

How do I check which version I am running Jonas? I copied the scripts from readme section mentioned at this URL https://github.com/derekantrican/GAS-ICS-Sync.

jonas0b1011001:
Open the copy of your script, at the bottom of Helpers.gs you can find a _thisVersion_ variable.

Did this error occur at the first run it did it start to appear after the script ran several times?

derekantrican:
Check near the bottom of Helpers.gs. You should see a `thisVersion = 5.2`. Don't worry if you don't have the latest version: version 5.3 was only released recently so it may have been after you got yours (and we don't have the best "update alerts" yet)

deanosauras:
Confirmed 5.3 and still occuring.
![image](https://user-images.githubusercontent.com/66968127/84698332-a081e080-af1d-11ea-8156-c124df7fc15e.png)

[...]

Issue #133, opened 2020-06-06
TypeError: Cannot read property 'dateTime' of undefined (line 493, file "Helpers")


javedrahman:
- Description of the problem - I am getting an error after upgrading to the new version.
- Screenshots 
![image](https://user-images.githubusercontent.com/11033953/83943762-66fef600-a7f6-11ea-81d2-bfc97fcb1de7.png)

- An input ics string that can reproduce the problem

You can see [this issue](https://github.com/derekantrican/GAS-ICS-Sync/issues/34) for an example of good information when submitting a bug

jonas0b1011001:
Seems like a duplicate of #115 , without additional info i can only suggest the same change as mentioned there.

Did the same ics file work with the previous version?

derekantrican:
@javedrahman Any update on this? What version are you running?

javedrahman:
Hi. I am on version 5.3. Instead of using the Install method, I have created a timed trigger for startSync to run every 15 minutes. It actually seems to sync just fine but throws the error at the end so overall the script shows as Failed in My Executions.

I think there is an issue with an item in my calendar. because I tried changing the date range for the ICS export and now it does not always fail.

jonas0b1011001:
Please try this solution [here](https://github.com/derekantrican/GAS-ICS-Sync/issues/135#issuecomment-644348397), this should fix your issue too. Unfortunately that's a common problem in 5.3.

derekantrican:
Seems like this is a duplicate of #135  so I'm going to close it as such

PR #151, opened 2020-09-12
Fix for TypeError: Cannot read property 'dateTime' of undefined bug

### LINKED ISSUE #115: TypeError: Cannot read property 'dateTime' of undefined
- Description of the problem

This just started happening with the script about a week ago. Until then, it was running normally. Now it is throwing an error for every iteration:

- Screenshots

4/10/20 3:53 AM | startSync | TypeError: Cannot read property "dateTime" from undefined. (line 488, file "Helpers") | time-based | 4/10/20 3:53 AM
-- | -- | -- | -- | --

- An input ics string that can reproduce the problem

I don't know which ICS entry is causing the error, but will share ICS link privately and reference this issue.

jonas0b1011001:
fixed #115 #133 #135 and #149 

Recommend to release as v5.4


derekantrican (approved):
Sorry for the late reply on this - had a busy day.

I've seen enough "yep, Jonas' version works for me" replies on #149 that I'll just go ahead and approve this. Just double-check that the merge was done correctly

Once you make the merge I'll publish a new version :)

PR #152, opened 2020-09-15
Change to use everyMinutes trigger for sync

### LINKED ISSUE #137: install() function not self-retriggering properly
[This is with v5.2]

After manually running the install() function once, it successfully triggers the startSync() function to run a few moments later, but the trigger to re-run the install() function again after howFrequent minutes happens only once, and that instance does not trigger startSync() or install() again. The syncing happens properly the one time (good logs, no errors, entries properly synced). The install() function does not output any logs.

i.e. in the logs I see, i sequence, install() run manually (by Editor), then startSync() run by trigger, then howFrequent minutes later install() run by trigger. And then nothing else after that.

Every few days I come back and manually run the install() function again, and the same pattern of executions as described above happens again. The expected continuous re-triggering of install() is not happening.

mastercko:
Instead of re-setting 'install' script every time, just set an "everyMinutes" trigger of the sync method. Fixes #137

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -90,13 +90,8 @@ function install(){
     //Delete any already existing triggers so we don't create excessive triggers
     deleteAllTriggers();
     
-    if (howFrequent < 1){
-      throw "[ERROR] \"howFrequent\" must be greater than 0.";
-    }
-    else{
-      ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create(); //Schedule sync routine to explicitly repeat
-      ScriptApp.newTrigger("startSync").timeBased().after(1000).create();//Start the sync routine
-    }
+    ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create(); //Schedule sync routine to explicitly repeat
+    ScriptApp.newTrigger("startSync").timeBased().after(1000).create();//Start the sync routine
   }catch(e){
     install();//Retry on error
   }
--- a/Code.gs
+++ b/Code.gs
@@ -94,7 +94,7 @@ function install(){
       throw "[ERROR] \"howFrequent\" must be greater than 0.";
     }
     else{
-      ScriptApp.newTrigger("install").timeBased().after(howFrequent * 60 * 1000).create();//Schedule next Execution
+      ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create(); //Schedule sync routine to explicitly repeat
       ScriptApp.newTrigger("startSync").timeBased().after(1000).create();//Start the sync routine
     }
   }catch(e){
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -1,3 +1,28 @@
+/**
+ * Takes an intended frequency in minutes and adjusts it to be the closest 
+ * acceptable value to use Google "everyMinutes" trigger setting (i.e. one of
+ * the following values: 1, 5, 10, 15, 30).
+ *
+ * @param {integer} The manually set frequency that the user intends to set
+ * @return {integer} The closest valid value to the intended frequency setting 
+ */
+function getValidTriggerFrequency(origFrequency) {
+  if (origFrequency == null) 
+    return 1; // if param is empty, just return 1, the lowest valid value.
+
+  
+  var adjFrequency = Math.round(origFrequency/5) * 5; // Set the number to be the closest divisible-by-5
+  adjFrequency = Math.max(adjFrequency, 1); // Make sure the number is at least 1 (0 is not valid for the trigger)
+  adjFrequency = Math.min(adjFrequency, 15); // Make sure the number is at most 15 (will check for the 30 value below)
+  
+  
+  if((adjFrequency == 15) && (Math.abs(origFrequency-30) < Math.abs(origFrequency-15)))
+    adjFrequency = 30; // If we adjusted to 15, but the original number is actually closer to 30, set it to 30 instead
+  
+  Logger.log("Intended frequency = "+origFrequency+", Adjusted frequency = "+adjFrequency);
+  return adjFrequency;
+}
+
 String.prototype.includes = function(phrase){ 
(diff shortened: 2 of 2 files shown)


derekantrican:
Thanks for your suggestion. We had this originally but it's restricted to certain inputs (eg 1, 5, 10, 15, etc) and our solution allowed any number you desired (eg run every 7 minutes) _and possibly some other improvements (can't remember)_. But #137 is definitely a problem (Google's problem, not ours) and stability may be more important. @jonas0b1011001 thoughts?

jonas0b1011001:
I don't know what the best option is, but waiting for google to fix the issue might be the worst.

We can go back to the 'old' method, add a setting to use the old method for those that have issues (as this seems to not happen to everyone) or add another check to see if install was successful (as already described [here](https://github.com/derekantrican/GAS-ICS-Sync/issues/137#issuecomment-666339558)).

mastercko:
Ah, I like the idea of having the flag in the settings for those of us having this problem but allowing the newer method for those that don't.

Ali1:
Yes please let's count on Google's recurrent triggers! Had problems like #137 and trying to figure out the cause is confusing.

Google trigger + a system for notifying fails will make this app much more dependable even if it means triggers are less frequent than desired.

Ali1:
Perhaps for people wanting to sync every 7 minutes, a LastRun conditional can be incorporated to enforce this while the Google trigger is everyMinute.

Helpers.gs
```diff
@@ -1,3 +1,28 @@
...
+ * Takes an intended frequency in minutes and adjusts it to be the closest 
+ * acceptable value to use Google "everyMinutes" trigger setting (i.e. one of
+ * the following values: 1, 5, 10, 15, 30).
+ *
+ * @param {integer} The manually set frequency that the user intends to set
+ * @return {integer} The closest valid value to the intended frequency setting 
+ */
+function getValidTriggerFrequency(origFrequency) {
+  if (origFrequency == null) 
+    return 1; // if param is empty, just return 1, the lowest valid value.
```
Merlin2001, line 11: To reduce server loads and also prevent issues caused by starting a sync while a long-running one is already in progress, I would suggest returning 15 here. It is also currently the default for new installs and seems a sensible default for most use cases.
(We're also not logging this case right now, should we?)

Merlin2001 (changes requested):

Helpers.gs
mastercko: This makes sense to me. I added a new commit that makes this change (and adds a log).

Merlin2001: Great! Now only the comment isn't accurate anymore 🙈

mastercko: haha, yeah, I saw those and fixed it in a couple of subsequent commits. Can you check again?

Merlin2001 (approved):
Perfect from my point of view! 👌 
(Unfortunately, I have no write access, either 🙈 )

Helpers.gs
```diff
@@ -1,3 +1,30 @@
...
+ * the following values: 1, 5, 10, 15, 30).
+ *
+ * @param {integer} The manually set frequency that the user intends to set. If blank, 15 is assumeed.
+ * @return {integer} The closest valid value to the intended frequency setting 
+ */
+function getValidTriggerFrequency(origFrequency) {
+  if (origFrequency == null) {
+    Logger.log("No intended frequency specified. Defaulting to 15 minutes.");
+    return 15; // if param is empty, just return the default value of 15.
+  }
```
jonas0b1011001, line 13: We are already checking validity of howFrequent in install().
I suggest removing the check in the install function and change this to:
```suggestion
 * @param {?integer} The manually set frequency that the user intends to set.
 * @return {integer} The closest valid value to the intended frequency setting. Defaulting to 15 if no valid input is provided.
 */
function getValidTriggerFrequency(origFrequency) {
  if (!origFrequency > 0) {
    Logger.log("No valid frequency specified. Defaulting to 15 minutes.");
    return 15;
  }
```

jonas0b1011001 (changes requested):

Code.gs
```diff
@@ -94,7 +94,7 @@ function install(){
       throw "[ERROR] \"howFrequent\" must be greater than 0.";
     }
     else{
-      ScriptApp.newTrigger("install").timeBased().after(howFrequent * 60 * 1000).create();//Schedule next Execution
+      ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create(); //Schedule sync routine to explicitly repeat
       ScriptApp.newTrigger("startSync").timeBased().after(1000).create();//Start the sync routine
     }
```
jonas0b1011001, line 99: With the previous change, the if statement is no longer needed at this point. Please remove it leaving only the trigger creations. After that i think we are good to go.

Helpers.gs
```diff
@@ -1,3 +1,30 @@
...
+ *
+ * @param {?integer} The manually set frequency that the user intends to set.
+ * @return {integer} The closest valid value to the intended frequency setting. Defaulting to 15 if no valid input is provided.
+ */
+function getValidTriggerFrequency(origFrequency) {
+  if (!origFrequency > 0) {
+    Logger.log("No valid frequency specified. Defaulting to 15 minutes.");
+    return 15;
+  }
+
```
jonas0b1011001, line 14: Please remove the excessive new line, helpers.gs is quite long already :)

Helpers.gs
```diff
@@ -1,3 +1,30 @@
...
+  
+  var adjFrequency = Math.round(origFrequency/5) * 5; // Set the number to be the closest divisible-by-5
+  adjFrequency = Math.max(adjFrequency, 1); // Make sure the number is at least 1 (0 is not valid for the trigger)
+  adjFrequency = Math.min(adjFrequency, 15); // Make sure the number is at most 15 (will check for the 30 value below)
+
```
jonas0b1011001, line 19: Please remove the excessive new line, helpers.gs is quite long already :)

jonas0b1011001 (changes requested):

mastercko (commented):
Updated with @jonas0b1011001 's requests.

jonas0b1011001 (approved):

PR #159, opened 2020-10-22
More robust periodic scheduling

eAndrius:
From personal experience the original approach (call install(), then schedule to call install() again in 15 minutes, repeat) is not immune to timezone / DST or wrong timezone settings. It would also fail at least once a week requiring manual intervention. Proposed approach uses Apps Script built-in periodic scheduler.


jonas0b1011001:
Thanks for the suggestion, we are already working on a similar PR #152 which will achieve the same.

Issue #155, opened 2020-10-01
Error, Retrying...Exception: Address unavailable


WoutVerschaeren:
When running the startSync function, I get this error 3 times before everything in the target calendar is deleted.

jonas0b1011001:
Please recheck the correct format of your sourceCalendars variable, and make sure the url is reachable.

There also seems to be a similar [issue](https://issuetracker.google.com/issues/36758958) on google's issuetracker for triggered runs.

WoutVerschaeren:
The url is definitely reachable, if I enter it in my browser, I can download and view it with no problems.

geogolem:
I was experiencing this when my ical links were down.. Sometimes the url is unavailable and in those cases, I just wanted to skip the syncing so that not all the entries were deleted.

I changed the section of code to skip if calendars are not available. I realize this doesn't actually correspond to the reason this issue was opened... but perhaps there should be some option to prevent deletion in case the ics url is unavailable for whatever reason. The little snippet of code I added below solves the issue in my case. I was going to open a new issue and submit a feature request but I'm not sure if I should or if others have a better way to handle this..

```
    //------------------------ Fetch URL items ------------------------
    var responses = fetchSourceCalendars(sourceCalendarURLs);
    //I added the next If block to handle cases when calendars are unavailable
    if (responses.length == 0){
      Logger.log("Skipping " + targetCalendarName);
      continue;
    }
    Logger.log("Syncing " + responses.length + " calendars to " + targetCalendarName);
```

Lonestarjeepin:
Duplicate of #343

PR #164, opened 2020-11-09
Skip (don't clear entries)if the calendar url address is unavailable,

geogolem:
Maybe some people have ics files on servers that are not always available.

If the url is not available, then don't delete all the entries - just skip that calendar.

This PR is not a solution to the original issue discussed here: https://github.com/derekantrican/GAS-ICS-Sync/issues/155
however, in my opinion it is a good idea, not to delete all the stuff if the server is unavailable. This creates an option *skipIfUrlUnavailable* . By default this is set to false - so the behavior is unchanged; however, if the user sets this to true, entries in a calendar will not be deleted if the url is not available.

I'm not sure if my implementation is ideal. It works for me, but it is likely that other will have suggestions to improve before this actually gets merged, if it ever does.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -24,6 +24,7 @@ var sourceCalendars = [                // The ics/ical urls that you want to get
   
 ];
 
+var skipIfUrlUnavailable = false;
 var howFrequent = 15;                  // What interval (minutes) to run this script on to check for new events
 var onlyFutureEvents = false;          // If you turn this to "true", past events will not be synced (this will also removed past events from the target calendar if removeEventsFromCalendar is true)
 var addEventsToCalendar = true;        // If you turn this to "false", you can check the log (View > Logs) to make sure your events are being read correctly before turning this on
@@ -137,7 +138,7 @@ function startSync(){
     var vevents;
     //------------------------ Fetch URL items ------------------------
     var responses = fetchSourceCalendars(sourceCalendarURLs);
-    if (responses.length == 0){
+    if (responses.length == 0 && skipIfUrlUnavailable){
       Logger.log("Skipping " + targetCalendarName);
       continue;
     }
--- a/Code.gs
+++ b/Code.gs
@@ -137,6 +137,10 @@ function startSync(){
     var vevents;
     //------------------------ Fetch URL items ------------------------
     var responses = fetchSourceCalendars(sourceCalendarURLs);
+    if (responses.length == 0){
+      Logger.log("Skipping " + targetCalendarName);
+      continue;
+    }
     Logger.log("Syncing " + responses.length + " calendars to " + targetCalendarName);
     
     //------------------------ Get target calendar information------------------------


derekantrican:
I think this makes sense to add. Go ahead and remove the variable option and add some more information to the log (something like "Source calendars unavailable. Skipping sync....")

geogolem:
> I think this makes sense to add. Go ahead and remove the variable option and add some more information to the log (something like "Source calendars unavailable. Skipping sync....")

As it is already the log looks like:
```
Nov 9, 2020, 9:40:25 PM	Info	Error, Retrying...Exception: Address unavailable: http://url-to-calendar
Nov 9, 2020, 9:40:28 PM	Info	Error, Retrying...Exception: Address unavailable: http://url-to-calendar
Nov 9, 2020, 9:40:31 PM	Info	Error, Retrying...Exception: Address unavailable: http://url-to-calendar
Nov 9, 2020, 9:40:31 PM	Info	Skipping TargetCalendarName
Nov 9, 2020, 9:40:40 PM	Info	Sync finished!
```

jonas0b1011001:
This will not prevent deletion if you sync multiple links to the same calendar and only some are unreachable.
We should either address this limitation or note it in the comments.

geogolem:
> This will not prevent deletion if you sync multiple links to the same calendar and only some are unreachable.
> We should either address this limitation or note it in the comments.

hmm i never considered that case, because I don't use it that way... but when I was looking through the code I was confused as I thought it overcomplicated things --> but now that you mention that case, I understand the existing code better.

Just quickly looking - it seems impossible to address this limitation without somehow tagging each event with the sourceUrls, then remove sourceUrl's if its no longer exists in that source url, but dont remove if that sourceUrl is unavailable.. Then only remove events if all sourceUrl tags are removed..

I dont like the sound of it since it sucks to have to tag events with sourceUrls.

Am I making sense - or do you see a better way to address this limitation?

jonas0b1011001:
> Just quickly looking - it seems impossible to address this limitation without somehow tagging each event with the sourceUrls, then remove sourceUrl's if its no longer exists in that source url, but dont remove if that sourceUrl is unavailable.. Then only remove events if all sourceUrl tags are removed..

That's pretty much the only way to handle it.

The issue with using addCalToTitle is that the Calendar Name can potentially change between two iterations.

There is actually no problem with checking response.length, it will be > 0 even if a calendar has 0 events.

geogolem:
> > Just quickly looking - it seems impossible to address this limitation without somehow tagging each event with the sourceUrls, then remove sourceUrl's if its no longer exists in that source url, but dont remove if that sourceUrl is unavailable.. Then only remove events if all sourceUrl tags are removed..
> 
> That's pretty much the only way to handle it.
> 
> The issue with using addCalToTitle is that the Calendar Name can potentially change between two iterations.
> 
> There is actually no problem with checking response.length, it will be > 0 even if a calendar has 0 events.

Well I think the best solution (for now) is to simply totally skip syncing to a targetCalendar if any 1 or more of its sourceUrls are unreachable and just log that one was unreachable and so a group of them were skipped. (only if removeEventsFromCalendar is enabled)

In other words if removeEventsFromCalendar is enabled and any of source calendar urls are unavailable for a specific targetCalendar behave as though none of the source calendar URLs are available and log appropriately.

geogolem:
The code has changed and there is a better solution now available. I will close this PR and create a new PR with updated code and new changes.

PR #165, opened 2020-11-11
Let installation errors bubble up so they're logged and fix callWithBackoff

Merlin2001:
Use the existing `callWithBackoff` method in `Helpers.gs` to limit retrying the script installation to 5 times, exponentially increasing the wait time between retries. This also gives us logging of errors during the installation.

Additionally, `callWithBackoff` is modified to do what it says on the tin and provide exponential backoff times (starting with 100ms, then 200ms, 400ms, etc.).

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -86,13 +86,12 @@ var email = "";                        // OPTIONAL: If "emailSummary" is set to 
 //!!!!!!!!!!!!!!!! DO NOT EDIT BELOW HERE UNLESS YOU REALLY KNOW WHAT YOU'RE DOING !!!!!!!!!!!!!!!!!!!!
 //=====================================================================================================
 function install(){
-  callWithBackoff(function() {
-    //Delete any already existing triggers so we don't create excessive triggers
-    deleteAllTriggers();
-    
-    ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create(); //Schedule sync routine to explicitly repeat
-    ScriptApp.newTrigger("startSync").timeBased().after(1000).create();//Start the sync routine
-  }, 5);  // Retry max. 5 times
+  //Delete any already existing triggers so we don't create excessive triggers
+  deleteAllTriggers();
+
+  //Schedule sync routine to explicitly repeat and schedule the initial sync
+  ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create();
+  ScriptApp.newTrigger("startSync").timeBased().after(1000).create();
 }
 
 function uninstall(){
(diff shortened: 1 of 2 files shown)


Code.gs
```diff
@@ -86,15 +86,13 @@ var email = "";                        // OPTIONAL: If "emailSummary" is set to
 //!!!!!!!!!!!!!!!! DO NOT EDIT BELOW HERE UNLESS YOU REALLY KNOW WHAT YOU'RE DOING !!!!!!!!!!!!!!!!!!!!
 //=====================================================================================================
 function install(){
-  try{
+  callWithBackoff(function() {
     //Delete any already existing triggers so we don't create excessive triggers
     deleteAllTriggers();
     
     ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create(); //Schedule sync routine to explicitly repeat
     ScriptApp.newTrigger("startSync").timeBased().after(1000).create();//Start the sync routine
```
derekantrican, line 94: The only problem with this is if line 93 succeeds, but line 94 is the failure, `callWithBackoff` could result in line 93 running multiple times until line 94 succeeds, leading to multiple triggers of `startSync` being created.

Have you seen problems with running the `install` function?

Merlin2001: Yes, I've had problems once, but that was due to me breaking `getValidTriggerFrequency` while playing around with it.

Your point is absolutely correct, and the current code behaves the same way, except that it will run much more often until the script is killed automatically (after 30 seconds, if I recall correctly).

There are two main intentions for this change 1) make `callWithBackoff` _actually_ exponential and 2) introduce logging of errors during `install`.

I would be absolutely fine with achieving 2) by just catching and logging errors and removing the retry entirely.

derekantrican: Actually, revising my concern for creating too many triggers: that should not happen since `deleteAllTriggers` is called first so it should clear any triggers created by the previous iteration of `callWithBackoff`.

I understand the reasoning to change the functionality of `callWithBackoff` itself - that seems fine.

derekantrican: I think now the only hesitations I have to using `callWithBackoff` is if we really need 5 retries on install. Maybe we just change the retry amount to `2`?

Merlin2001: Oh, you're absolutely right. It's late 😅   
I'd be completely on board with reducing the retries to 2, but maybe even retrying once (after waiting only 100ms) might be unnecessary. What about we just remove all exception catching from `install()`?

In my opinion this would have the following benefits:  
a) Exceptions are automatically logged prominently in red (since they are not handled).
b) The function actually shows up as _Failed_ in _My Executions_ giving the user immediate feedback that something went wrong.

Especially b) is something that's currently hidden well by handling the errors internally.

So my new suggestion would simply be:
```javascript
function install(){
  //Delete any already existing triggers so we don't create excessive triggers
  deleteAllTriggers();
  
  //Schedule sync routine to explicitly repeat and schedule the initial sync
  ScriptApp.newTrigger("startSync").timeBased().everyMinutes(getValidTriggerFrequency(howFrequent)).create();
  ScriptApp.newTrigger("startSync").timeBased().after(1000).create();
}
```

derekantrican: Sure, sounds good to me! I think before when we were calling `install` on every iteration to trigger the next iteration it was more important. But since we recently reverted the trigger functionality I totally agree with you

derekantrican (approved):

PR #166, opened 2020-11-16
Fix bug with recurring events and multiple calendars.

jma127:
The `recurringEvents` array was not being reset on each new calendar.
This would cause "calendar contamination" when using this script with
more than one Google Calendar, as any recurring events processed on the
first calendar would also end up added to the second calendar.

Out of an abundance of caution, I've added a section of code in the main
loop that resets all "per-calendar" global variables. Longer-term, we
might think about how to factor the global variables out entirely.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -98,8 +98,9 @@ function uninstall(){
   deleteAllTriggers();
 }
 
-var targetCalendarId;
-var targetCalendarName;
+var startUpdateTime;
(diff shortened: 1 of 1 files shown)


derekantrican:
I think it seems fine to me. @jonas0b1011001 to double-check

jonas0b1011001 (approved):

PR #178, opened 2021-01-18
Update getLastestVersion function

ruggi99:
Update getLatestVersion function to use GitHub APIs.
I think it's better then parse HTML with RegExps.
Let me know


derekantrican (approved):
Seems like a good plan!

PR #188, opened 2021-02-27
several fixes for reported issues

jonas0b1011001:
Ignore unsupported 'url' properties
More robust targetCalendar selection
Handle RDates when onlyFutureEvents = true
Improved Recurrence Exception Handling 
Add Backoff to cleanup procedure
Add support for Events without uid

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -574,15 +574,12 @@ function processEventCleanup(){
       
       if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
         Logger.log("Deleting old event " + currentID);
-        try{
+        callWithBackoff(function(){
           Calendar.Events.remove(targetCalendarId, calendarEvents[i].id);
           if (emailSummary){
             removedEvents.push([[calendarEvents[i].summary, calendarEvents[i].start.date||calendarEvents[i].start.dateTime], targetCalendarName]);
           }
-        }
-        catch (err){
-          Logger.log(err);
-        }
+        }, 2);
       }
     }
 }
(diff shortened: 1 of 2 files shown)


Helpers.gs
```diff
@@ -502,43 +527,41 @@ function checkSkipEvent(event, icalEvent){
...
+      Logger.log("Updating existing event instance");
+      Calendar.Events.update(recEvent, targetCalendarId, eventInstanceToPatch[0].id);
     }
     catch(error){
       Logger.log(error); 
     }
   }
   else{
+    Logger.log("No Instance matched, adding as new event!");
     try{
```
derekantrican, line 563: Don't know if we also want to replace this and the `try` above (for `Events.update`) (and maybe more) with `callWithBackoff` like you did below

Helpers.gs
```diff
@@ -108,7 +108,8 @@ function fetchSourceCalendars(sourceCalendarURLs){
  */
 function setupTargetCalendar(targetCalendarName){
   var targetCalendar = Calendar.CalendarList.list().items.filter(function(cal) {
-    return cal.summary == targetCalendarName;
+    var name = cal.summaryOverride || cal.summary;
+    return name == targetCalendarName && cal.accessRole == "owner";
```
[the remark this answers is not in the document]
jonas0b1011001: True, _writer_ should work too. _freeBusyReader_ and _reader_ should be ignored.

Helpers.gs
jonas0b1011001: I guess we can add it to all api calls.

derekantrican:
@jonas0b1011001 ping on this - if you have a chance, I'd like to pull these in and release a new version. Let me know if you want any clarification on my comments

jonas0b1011001:
Done, sorry for the delay.

derekantrican (approved):
No worries. Thanks for updating it!

PR #192, opened 2021-03-23
HTML Support

Ali1:
Tested working, description showing gracefully on Google Calendar interface with HTML formatting displayed.

However clients that may be syncing to Google Calendar (i.e. iOS native Calendar) do not support HTML and just display the raw HTML code as the event description.

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -324,8 +324,10 @@ function createEvent(event, calendarTz){
     newEvent.summary = "(" + calName + ") " + newEvent.summary;
   }
   
-  if (event.hasProperty('description'))
+  if (event.hasProperty('description')) {
     newEvent.description = icalEvent.description;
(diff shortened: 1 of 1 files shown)


Ali1:
Sorry ignore the first 2 commits, how do you feel about HTML support (the 3rd commit)

derekantrican:
Couple things:

- If we should "ignore the first two commits", then please revert the changes there or start a new PR with just the suggested changes
- Can you give an example of what this accomplishes? Screenshots, example ICS url, etc?

Ali1:
I've decided to go a different way and just keep things plaintext in my upstream app.

The use case would be syncing with any ICS file that has HTML descriptions (my app was adding evernote notes in HTML to calendar events).

For reference if anyone is searching in the future: https://github.com/Ali1/GAS-ICS-Sync/commit/9da2f0206feb00cfaeaccf0b261911421ebfefe8

Issue #196, opened 2021-04-13
Summary email doesn't include all event related changes


bandtank:
To work around the issue in #185, I copied the latest version of each file from the repo into my Google Scripts project. It seems to work, but the summary email didn't accurately report every change.

Here is the summary:

```
GAS-ICS-Sync made the following changes to your calendar:

Test Calendar 4: 1 added events
    Cancel Amazon Prime at 2021-05-12T08:30:00-06:00
```

The event shown above was expected, so that part is fine. However, other changes were made according to the log. More importantly, more changes were _expected_ and the resulting calendars were correctly updated.

```
Apr 13, 2021, 1:44:15 PM	Info	Result: 1
Apr 13, 2021, 1:44:15 PM	Info	Syncing 1 calendars to Test Calendar 1
Apr 13, 2021, 1:44:15 PM	Info	Working on calendar: <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Fetched 37 existing events from Test Calendar 1
Apr 13, 2021, 1:44:15 PM	Info	Parsed 39 events from ical sources
Apr 13, 2021, 1:44:15 PM	Info	Processing 39 events
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
Apr 13, 2021, 1:44:15 PM	Info	Skipping unchanged Event <redacted>
[...]

Issue #123, opened 2020-04-23
Scope requests might be to permissive


chregon:
I hope this is the right place for this, if not please inform me.

I think many users find the requested scope is a bit too much (e.g. read, edit, and delete all emails).

In my case, I have the script running with only:
```json
"oauthScopes": [
    "https://www.googleapis.com/auth/calendar",
    "https://www.googleapis.com/auth/script.external_request",
    "https://www.googleapis.com/auth/script.scriptapp"
  ],
```
in the appscript.json. So no emails, and no task integration, but I bet many users don't find these necessary.

I think having the scope set manually, and maybe even informing the user of how to remove unneeded scope can help reach more people on the fence of using this. Also, might prevent some issues like [this issue](https://github.com/derekantrican/GAS-ICS-Sync/issues/118) from becoming catastrophic down the line.

See, [Authorization Scopes documentation](https://developers.google.com/apps-script/concepts/scopes#setting_explicit_scopes) for more.

jonas0b1011001:
Hi,
thank you for the feedback! I agree that the requested permission might be daunting to some users, but as this is open source nobody has to blindly trust the script to not misuse the permissions.

> (e.g. read, edit, and delete all emails)

This will be changed in the next version by switching from _GMailApp_ to _MailApp_.
[Unlike GmailApp, MailApp's sole purpose is sending email. MailApp cannot access a user's Gmail inbox.](https://developers.google.com/apps-script/reference/mail/mail-app)

> So no emails, and no task integration, but I bet many users don't find these necessary.

I see that some features are not used by some users, the best solution would be to dynamically request permissions based on the settings activated by the user but this is as far as i know currently not possible.

> I think having the scope set manually, and maybe even informing the user of how to remove unneeded scope

As "[Apps Script automatically detects what scopes are needed](https://developers.google.com/apps-script/concepts/scopes)" the script will always request all permissions that _might_ be used by the script. An explaining wiki post about Permissions and how to adjust them is a good idea.

> Also, might prevent some issues like this issue from becoming catastrophic down the line.

The script will **never** edit/delete events that were not originally created by it!

derekantrican:
[...]

PR #197, opened 2021-04-14
Beginning to refactor globals into locals

bandtank:
In response to #196, I started refactoring the code to remove several globals. I also removed some unnecessary conditional logic and optimized a handful of temporary variables. Gating logic has been moved into the email summary function, which also removed the need for the boolean. Lastly, I tweaked a few comments to be more consistent with the rest of the comments.

I tested the changes using 12 calendars in a variety of configurations along with added, modified, and removed events.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -42,7 +42,9 @@ var addAttendees = false;              // Whether to add the attendee list. If t
 var defaultAllDayReminder = -1;        // Default reminder for all day events in minutes before the day of the event (-1 = no reminder, the value has to be between 0 and 40320)
                                        // See https://github.com/derekantrican/GAS-ICS-Sync/issues/75 for why this is neccessary.
 var addTasks = false;                  // Add tasks from source calendars to target calendars
-var email = "";                        // OPTIONAL: If provided, a summary email will be sent if anything changes
+var enableUpdateEmail = true;          // Send an email if a new version of the script is available
+var enableSummaryEmail = false;        // Send a summary email after each run if anything changes
+var email = "";                        // To receive emails, an address must be provided
 
 /*
 *=========================================
@@ -124,7 +126,8 @@ function startSync(){
   
   PropertiesService.getScriptProperties().setProperty('LastRun', new Date().getTime());
   
-  checkForUpdate();
+  if(enableUpdateEmail)
+    checkForUpdate(email);
   
   if (onlyFutureEvents)
     startUpdateTime = new ICAL.Time.fromJSDate(new Date());
@@ -211,7 +214,8 @@ function startSync(){
     }
   }
   
-  sendSummary(email, addedEvents, modifiedEvents, removedEvents);
+  if(enableSummaryEmail)
+    sendSummary(email, addedEvents, modifiedEvents, removedEvents);
 
   Logger.log("Sync finished!");
   PropertiesService.getScriptProperties().setProperty('LastRun', 0);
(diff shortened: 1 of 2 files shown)


Code.gs
```diff
@@ -130,20 +129,17 @@ function startSync(){
...
+  for (var calendar of condenseCalendarMap(sourceCalendars)){
     //------------------------ Reset globals ------------------------
     calendarEvents = [];
     calendarEventsIds = [];
     icsEventsIds = [];
     calendarEventsMD5s = [];
     recurringEvents = [];
-    addedEvents = [];
-    modifiedEvents = [];
-    removedEvents = [];
```
derekantrican, line 146: Does this change mean that `addedEvents`, etc is now relevant to all calendars and not calendar-specific?

derekantrican: Oh, lower I think I see that you're just adding `condenseCalendarMap` to references instead

Helpers.gs
```diff
@@ -778,33 +776,37 @@ function parseNotificationTime(notificationString){
 /**
 * Sends an email summary with added/modified/deleted events.
 */            
-function sendSummary() {
-  var subject;
-  var body;
+function sendSummary(email, addedEvents, modifiedEvents, removedEvents) {
+  if (email == "") {
+    Logger.log("A summary email will not be sent because an email address has not been provided.")
```
derekantrican, line 781: I would like to avoid adding extra log lines as a line like this could seem like an error to a user. It would make sense to have this if you kept the `emailSummary` bool (since the user *requested* an email, but didn't provide an email address), but if you're removing that bool, let's also remove this log line

derekantrican:
I added a couple comments, but otherwise I think it looks pretty good. @jonas0b1011001 to double-check functionality and see if he wants to add any comments as well

Helpers.gs
bandtank: Ok, fair enough. I prefer verbosity so people never have to guess why things did or did not happen. Since it's a log file, it makes sense to me to emit messages when decisions are made.

Code.gs
bandtank: In a nutshell, I removed the lines that modify the original arrays to prevent unintended changes or consequences in other locations. `addedEvents` was aggregated events from every calendar and that behavior hasn't changed. The difference is that the various events aggregators are not global and should be easier to pass around without causing bugs like #196.

Helpers.gs
derekantrican: The new GAS UI also shows log messages at the bottom when running (rather than a separate window) so users are paying attention to them more. We've gotten more comments about the "Timezone _____ unsupported! Using ____ instead" and whether that was an error, something will mess up their calendar, etc

jonas0b1011001:
Looks fine to me.
The only thing we are losing is the ability to enable "new version available" mails without receiving summary mails aswell. Not sure if we should keep the extra switch.

bandtank:
> Looks fine to me.
> The only thing we are losing is the ability to enable "new version available" mails without receiving summary mails aswell. Not sure if we should keep the extra switch.

I think it would be better to keep the email functions separate from each other. `sendSummary` has a specific purpose and the name makes it obvious. The alert for a new version would make sense to put in a separate function, which is essentially the way it currently is. The removal of the email boolean didn't impact the transmission of an update email. `checkForUpdate` is dispatching the email after checking for the existence of an address just like the modified version of `sendSummary` in this PR.

An alternative method would be to parameterize a generic `sendMail` function, but the `Mailapp.sendEmail` call is already really simple. Wrapper functions are probably the most maintainable way to organize email transmission functions.

jonas0b1011001:
That's what i was thinking. I'm happy with your example.

derekantrican:
I think one thing helpful to note here is that you could only use the var as a bool and use `Session.getActiveUser().getEmail()` to get the email address. Though I don't know if that would cause more perm requests (related to #123)

bandtank:
I think it's better to make more of it as automated as possible, but the users may want to send the notifications to another email, so a way to choose would be good. I'm going to submit a different PR with a few tweaks to the notification flow, which I didn't think were a good fit for this PR due to the more limited scope.

PR #199, opened 2021-04-28
Moved ICAL code

rufuswilson:
I moved some code relating to the ICAL event from processEvent to createEvent


derekantrican (approved):
Seems fine

PR #200, opened 2021-04-28
Adding option to change colour of events

rufuswilson:
I am synchronising my events in a pre-existing calendar and being able to change the colour of the events allows me to better separate imported events and other events.


derekantrican:
Similar to #204 I'm going to close this as "too specific to your situation"

PR #201, opened 2021-04-28
Default event visibility

rufuswilson:
Some of my events had a visibility class of private and I wanted to be able to override it to the calendar setting.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -41,7 +41,7 @@ var addCalToTitle = false;             // Whether to add the source calendar to 
 var addAttendees = false;              // Whether to add the attendee list. If true, duplicate events will be automatically added to the attendees' calendar.
 var defaultAllDayReminder = -1;        // Default reminder for all day events in minutes before the day of the event (-1 = no reminder, the value has to be between 0 and 40320)
                                        // See https://github.com/derekantrican/GAS-ICS-Sync/issues/75 for why this is neccessary.
-var forceVisibility = "default";       // Changes the visibility of the event ("default", "public", "private"). Anything else will revert to the class value of the ICAL event.
+var forceVisibility = false;           // Changes the visibility of the event ("default", "public", "private"). Anything else will revert to the class value of the ICAL event.
 var addTasks = false;
 
 var emailSummary = false;              // Will email you when an event is added/modified/removed to your calendar
--- a/Code.gs
+++ b/Code.gs
@@ -41,7 +41,7 @@ var addCalToTitle = false;             // Whether to add the source calendar to 
 var addAttendees = false;              // Whether to add the attendee list. If true, duplicate events will be automatically added to the attendees' calendar.
 var defaultAllDayReminder = -1;        // Default reminder for all day events in minutes before the day of the event (-1 = no reminder, the value has to be between 0 and 40320)
                                        // See https://github.com/derekantrican/GAS-ICS-Sync/issues/75 for why this is neccessary.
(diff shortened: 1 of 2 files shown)


Code.gs
```diff
@@ -41,6 +41,7 @@ var addCalToTitle = false;             // Whether to add the source calendar to
 var addAttendees = false;              // Whether to add the attendee list. If true, duplicate events will be automatically added to the attendees' calendar.
 var defaultAllDayReminder = -1;        // Default reminder for all day events in minutes before the day of the event (-1 = no reminder, the value has to be between 0 and 40320)
                                        // See https://github.com/derekantrican/GAS-ICS-Sync/issues/75 for why this is neccessary.
+var forceVisibility = false;           // Changes the visibility of the event ("default", "public", "private"). Anything else will revert to the class value of the ICAL event.
```
derekantrican, line 44: Let's call this `override` rather than `force`

Helpers.gs
```diff
@@ -355,10 +355,14 @@ function createEvent(event, calendarTz){
   if (event.hasProperty('location'))
     newEvent.location = icalEvent.location;
 
-  if (event.hasProperty('class')){
-    var classString = event.getFirstPropertyValue('class').toString().toLowerCase();
-    if (["default", "public", "private", "confidential"].indexOf(classString) > -1)
-      newEvent.visibility = classString;
+  if ( ["default", "public", "private"].includes(forceVisibility.toLowerCase()) ) {
```
derekantrican, line 358: Why does this array not include `confidential` like the array on 363? I think better to add that value and break out the array to something like `var validVisibilityValues` and use it in both places

Helpers.gs
```diff
@@ -355,10 +355,14 @@ function createEvent(event, calendarTz){
...
+    newEvent.visibility = forceVisibility;
+  } else {
+    if (event.hasProperty('class')){
```
derekantrican, line 361: This `if` can be combined with the `else` above (ie `else if`). No need to have them separate.

derekantrican (changes requested):

Code.gs
rufuswilson: Agreed. Changes committed.

Helpers.gs
rufuswilson: Agreed. Changes committed.

Helpers.gs
rufuswilson: agreed and done

Code.gs
```diff
@@ -41,6 +41,7 @@ var addCalToTitle = false;             // Whether to add the source calendar to
 var addAttendees = false;              // Whether to add the attendee list. If true, duplicate events will be automatically added to the attendees' calendar.
 var defaultAllDayReminder = -1;        // Default reminder for all day events in minutes before the day of the event (-1 = no reminder, the value has to be between 0 and 40320)
                                        // See https://github.com/derekantrican/GAS-ICS-Sync/issues/75 for why this is neccessary.
+var overrideVisibility = false;           // Changes the visibility of the event ("default", "public", "private"). Anything else will revert to the class value of the ICAL event.
```
derekantrican, line 44: Thanks for updating this. Even though it doesn't matter much from the perspective of someone familiar with Javascript, let's change the default value from `false` to a string (like `""` you had before) to make it more obvious to users. I feel like if it should be set to a string value, the default value should probably also be a string

rufuswilson: OK

derekantrican (approved):

PR #203, opened 2021-04-28
Improvement to callWithBackoff

rufuswilson:
In some cases, I had some errors and callWithBackoff was trying obstinately even if it was just not possible.
One such case was Quota exceeded.

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -834,7 +834,7 @@ function sendSummary() {
 function callWithBackoff(func, maxRetries) {
   var tries = 0;
   var result;
-  while(true){
+  while ( tries > maxRetries ) {
     tries++;
     try{
       result = func();
@@ -847,7 +847,7 @@ function callWithBackoff(func, maxRetries) {
         throw e;
       } else if ( e.message.includes("Forbidden") ) {
         return null;
-      } else if ( maxRetries < tries ) {
+      } else if ( tries > maxRetries ) {
         Logger.log(`Error, giving up after trying ${maxRetries} times [${e}]`);
         return null;
       } else {
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -831,6 +831,17 @@ function sendSummary() {
  * @param {Number} maxRetries - How many times the function should try if it fails
  * @return {?Calendar.Event} The Calendar.Event that was added in the calendar, null if func did not complete successfully
  */
+var backoffRecoverableErrors = [
+  "Exception: Service invoked too many times in a short time",
+  "Exception: Rate Limit Exceeded",
+  "Exception: Quota Error: User Rate Limit Exceeded",
+  "Service error: Spreadsheets",
+  "Exception: User rate limit exceeded",
+  "Exception: Internal error. Please try again.",
+  "Exception: Cannot execute AddColumn because another task",
+  "Service invoked too many times in a short time:",
+  "Exception: Internal error.",
+  "Exception: Limit Exceeded: DriveApp."];
 function callWithBackoff(func, maxRetries) {
   var tries = 0;
   var result;
@@ -840,19 +851,20 @@ function callWithBackoff(func, maxRetries) {
       result = func();
       return result;
     }
-    catch(e){
-      if ( e.message.includes("Rate Limit Exceeded") ) {
-        throw e;
-      } else if ( e.message.includes("is not a function") ) {
-        throw e;
-      } else if ( e.message.includes("Forbidden") ) {
+    catch(err){
+      if ( err.message.includes("is not a function")  || !backoffRecoverableErrors.some(function(e){
+              return  err.message.toString().slice(0,e.length) == e  ;
+            }); ) {
+        throw err;
+      } else if ( err.message.includes("Forbidden") ) {
         return null;
-      } else if ( tries > maxRetries ) {
-        Logger.log(`Error, giving up after trying ${maxRetries} times [${e}]`);
+      } else if ( tries > maxRetries) {
+        Logger.log(`Error, giving up after trying ${maxRetries} times [${err}]`);
         return null;
       } else {
-        Logger.log( "Error, Retrying..." + e );
-        Utilities.sleep((2**tries) * 100);
+        Logger.log( "Error, Retrying..." + err );
+        Utilities.sleep (Math.pow(2,tries)*100) + 
+                            (Math.round(Math.random() * 100));
       }
     }
   }
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -832,9 +832,9 @@ function sendSummary() {
  * @return {?Calendar.Event} The Calendar.Event that was added in the calendar, null if func did not complete successfully
  */
 var backoffRecoverableErrors = [
-  "Service invoked too many times in a short time",
-  "Rate Limit Exceeded",
-  "Internal error"];
+  "service invoked too many times in a short time",
+  "rate limit exceeded",
+  "internal error"];
 function callWithBackoff(func, maxRetries) {
   var tries = 0;
   var result;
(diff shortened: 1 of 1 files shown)


Helpers.gs
```diff
@@ -833,26 +833,29 @@ function sendSummary() {
...
       return result;
     }
     catch(e){
-      if (tries < maxRetries){
-        Logger.log(`Error, retrying in ${timeout}ms... [${e}]`);
-        Utilities.sleep(timeout);
-        timeout*=2; // Exponentially increase timeout
-      }
-      else {
+      if ( e.message.includes("Rate Limit Exceeded") ) {
```
derekantrican, line 844: Is there a reason "Rate Limit Exceeded" immediately throws rather than retrying?

Helpers.gs
```diff
@@ -833,26 +833,29 @@ function sendSummary() {
  */
 function callWithBackoff(func, maxRetries) {
   var tries = 0;
-  var timeout = 100;
   var result;
-  while(tries < maxRetries){
+  while(true){
```
derekantrican, line 837: Let's keep the condition here. In case something unintended/unexpected happens with the below code I'd rather it still hit this condition rather than become an infinite loop

Helpers.gs
```diff
@@ -833,26 +833,29 @@ function sendSummary() {
...
-        timeout*=2; // Exponentially increase timeout
-      }
-      else {
+      if ( e.message.includes("Rate Limit Exceeded") ) {
+        throw e;
+      } else if ( e.message.includes("is not a function") ) {
+        throw e;
+      } else if ( e.message.includes("Forbidden") ) {
+        return null;
+      } else if ( maxRetries < tries ) {
```
derekantrican, line 850: This is a nitpick, but let's reverse this condition for readability: `tries > maxRetries`

derekantrican (changes requested):

Helpers.gs
rufuswilson: The quotas are defined per day. In my opinion, it doesn't make sense to try again right away.

Helpers.gs
rufuswilson: Yes, makes sense.

Helpers.gs
rufuswilson: OK.

Helpers.gs
derekantrican: Here's a good reference I found online: https://www.oreilly.com/library/view/going-gas/9781491940457/ch04.html In the section "Code for exponential backoff" they have a list of backoff error messages. "Rate Limit Exceeded" and "User Reate Limit Exceeded" are for the quota dealing with speed quotas and not daily quotas.

rufuswilson: Yes, you're right. I adapted the code to take that into account. I also added the random delay from your reference.

Helpers.gs
```diff
@@ -831,28 +831,43 @@ function sendSummary() {
...
-      if (tries < maxRetries){
-        Logger.log(`Error, retrying in ${timeout}ms... [${e}]`);
-        Utilities.sleep(timeout);
-        timeout*=2; // Exponentially increase timeout
-      }
-      else {
-        Logger.log(`Error, giving up after trying ${maxRetries} times [${e}]`);
+    catch(err){
+      if ( err.message.includes("is not a function")  || !backoffRecoverableErrors.some(function(e){
+              return  err.message.toString().slice(0,e.length) == e  ;
```
derekantrican, line 856: So I have to admit that I don't fully understand what this is doing (js isn't my strongest language), but I think that the essence of the `some.....slice.....` section is "if any of `backoffRecoverableErrors` is found in `err.message` (starting with the beginning - hence `slice(0`)". Is that how you would interpret it too?

A couple things I've noticed. First, here you pass in `err.message` to the `backoffRecoverableErrors.some` callback function and the example site I shared just uses `err`. Not sure if that makes a difference, but it might be what causes the "Exception: " prefix. If that's the case, then this `some` function would never match.

To solve this issue (and possibly condense our list a bit) can we instead trim it to something like:

```javascript
var backoffRecoverableErrors = [
  "Service invoked too many times in a short time",
  "Rate Limit Exceeded",
  "Internal error",
  "Service invoked too many times in a short time:"];
...
            if ( err.message.includes("is not a function")  || !backoffRecoverableErrors.some(function(e){
              return new RegExp(e, 'i').test(err.toString())
            }) ) {
...
```

This would limit our array a lot (I also removed ones that don't really apply to our script - like Drive, Spreadsheets, etc). Thoughts?

rufuswilson: Yes exactly. It is an equivalent of an startWith function looping over all the error messages. 

You are correct about the never matching. I misinterpreted the name of the variable.

Good idea about the condensing.

Helpers.gs
```diff
@@ -831,28 +831,36 @@ function sendSummary() {
...
-      else {
-        Logger.log(`Error, giving up after trying ${maxRetries} times [${e}]`);
+    catch(err){
+      if ( err.message.includes("is not a function")  || !backoffRecoverableErrors.some(function(e){
+              return new RegExp(e, 'i').test(err.toString())
+            }) ) {
+        throw err;
+      } else if ( err.message.includes("Forbidden") ) {
+        return null;
+      } else if ( tries > maxRetries) {
```
jonas0b1011001, line 854: That part is unreachable, should be =

Helpers.gs
```diff
@@ -831,28 +831,36 @@ function sendSummary() {
...
+var backoffRecoverableErrors = [
+  "Service invoked too many times in a short time",
+  "Rate Limit Exceeded",
+  "Internal error"];
 function callWithBackoff(func, maxRetries) {
   var tries = 0;
-  var timeout = 100;
   var result;
-  while(tries < maxRetries){
+  while ( tries < maxRetries ) {
```
jonas0b1011001, line 841: This has been wrong before but i think this should be _<=_. With maxRetries set to 0 i would expect the _func_ to get called once but not again if it fails. As of now it would not be called at all.

Helpers.gs
jonas0b1011001: If we implement the change from my other comment, this is obsolet :)

Helpers.gs
derekantrican: I'm worried about the speed costs of my solution, though. Constructing a regex object for EVERY exception message in EVERY backoff exception for EVERY api call seems quite costly. And doing a regex match is probably also slow when compared to `includes`.

Maybe if we make the whole array lower/upper case and use a callback like `e.includes(err.message.toString().toLowerCase()`?

Thoughts? @jonas0b1011001 ?

rufuswilson: I see your point. It does make sense. Will wait for inputs from JS experts to commit changes.

Helpers.gs
rufuswilson: Agreed. Change committed

Helpers.gs
derekantrican: Benchmark tests: turns out `indexOf` is crazy fast, but `includes` is still ~8 times faster than regex

derekantrican: So let's do

```javascript
var backoffRecoverableErrors = [
  "service invoked too many times in a short time",
  "rate limit exceeded",
  "internal error"];
...
            if ( err.message.includes("is not a function")  || !backoffRecoverableErrors.some(function(e){
              return err.message.toLowerCase().indexOf(e) > -1;
            }) ) {
...
```

rufuswilson: Perfect. Change committed

derekantrican: @rufuswilson one final tweak - the strings in the `backoffRecoverableErrors` array will have to be lowercase as well

rufuswilson: Indeed

derekantrican (approved):

PR #204, opened 2021-04-28
Ignore all d ay events

rufuswilson:
Some of my calendars contain all day events for bank holidays which makes a mess with several duplicate events.


derekantrican:
I think this is a little too specific to your situation so I'm going to skip on this one

PR #205, opened 2021-04-28
Changed to UserProperties

rufuswilson:
I share the project between different accounts and it happened that the script did not run for one user as it was already running for another.


derekantrican (approved):
Makes sense

PR #208, opened 2021-05-27
Reset global variables

rufuswilson:
Global variables that should not be reset between when processing each calendars


derekantrican:
This is to make the summary email include added/modified/deleted events for all calendars, correct? Based on what I'm seeing, looks like you're fixing an issue where it would just have the data for the last calendar

rufuswilson:
Yes exactly, you're right





---------------------------------------------------
Tchô

derekantrican (approved):

PR #213, opened 2021-06-09
Update tzid.gs for Webex time

jasonwinget:
Webex does not follow standard Windows Timezone names in Outlook (does not use "Standard" in name, i.e. "Pacific Standard Time" is only "Pacific Time")


derekantrican (commented):
Sounds good. Don't know anything about webex calendars, but looks like you know more than I do. Looks like you've covered only US timezones - are there other webex-specific timezones we could include in the list?

PR #214, opened 2021-06-12
no add alert reverts to calendar's default alert

rufuswilson:
When addAlert is set to false, the event will have the default alerts of the calendar.
I would have expected that it would not have any alerts at all.


derekantrican:
I understand the confusion, though I feel like this could still be a desired behavior. Maybe we should treat it more as an enum (addAlerts: yes/no/default)? @jonas0b1011001 thoughts?

rufuswilson:
I actually very much like the idea

derekantrican:
Seems fine to me. If Jonas doesn't have any thoughts I'll merge it in

jonas0b1011001:
I'm happy with the changes

rufuswilson:
Just before you merge them, I would like to test the default option. Just
realized I didn't test it yet





---------------------------------------------------
Tchô

jonas0b1011001:
> When addAlert is set to false, the event will have the default alerts of the calendar.

So seems like we could go with _doing nothing_ as the default option

rufuswilson:
By “doing nothing” you mean using the calendar’s default notifications or not having any notifications at all?



---------------------------------------------------
Tchô

rufuswilson:
Finished the testing and all options behave as they should.

derekantrican:
I think the changes seem fine to me. Glad you've also got the `default: ` in your switch-case (and I think "target calendar default" is a fine default)

PR #217, opened 2021-06-18
Update Code.gs > getPastDaysIfOnlyFutureEvents

la32094uf:
Found a way to get past events, if onlyFutureEvents is set to true. Great for big calendars, which lead to timeouts. With this tweak, it is still possible to see some events in the past.


ryancasler:
I was adding a check that it was greater than 0 to allow folks to still sync only future events when it occurred to me that as long as the number is 0, you'd get the same result.  DER!   (I'm a coding novice at best so I'm glad it occurred to me before posting it.)

But I did want to say, thanks for this!!!   I have noticed recently that I am getting a LOT of notifications about my script timing out.  My calendar is just about 2 years old and I have at least 4-5 events per day.  I still want those old ones to be shown for reference for a little while.  This is a great compromise.  LOVE IT!

PR #235, opened 2021-10-17
Currently the script fails execution if it encounters a sourceCalendar URl that is not available.

geogolem:
This commit logs the fact that the url is available
and that the syncing of the targetcalendar will be skipped
and execution is allowed to continue for syncing subsequent calendars etc.

This commit is sort of a modified implementation of:
https://github.com/derekantrican/GAS-ICS-Sync/pull/164


derekantrican:
So the method is skip all syncing to that Google Calendar if any of the source calendars are unavailable? I'm not sure I like that solution - what if I have 3 ical sources for 1 Google calendar, but 1 is unavailable? Then I would stop getting updates from those other calendars.

geogolem:
> So the method is skip all syncing to that Google Calendar if any of the source calendars are unavailable? I'm not sure I like that solution - what if I have 3 ical sources for 1 Google calendar, but 1 is unavailable? Then I would stop getting updates from those other calendars.

We can probably change the functionality to handle different cases --> right now, the scirpt just dies at that point the moment it finds a url unavailable. My little change makes it keep going since other url's correspoinding to other targets might be available.

Lonestarjeepin:
Closing per conversation with Jonas.  This fix was pushed to master via #245 .

PR #239, opened 2021-11-08
Fix bug with rrules and n occurrences

rufuswilson:
I noticed that If I have a recurring event which stops after n occurrences. The number of occurrences does not reduce with time but they are shifted.

I tested Monthly and Daily recurrences but not Weekly nor Yearly. I added support for INTERVAL as well.

To reproduce simply create a daily recurring event starting the day before but a bit later in the day than the current time and finishing after 4 occurrences. Run the script and you'll still see 4 events but starting one day later.

(I have the feeling this could be simplified a lot if the ICAL.RecurExpansion class could provide the total number of occurrences for time-limited recurrences)

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -486,12 +486,53 @@ function checkSkipEvent(event, icalEvent){
       
       if (newStartDate != null){//At least one instance is in the future
         newStartDate.timezone = icalEvent.startDate.timezone;
+        var oldStartDate = icalEvent.startDate;
         var diff = newStartDate.subtractDate(icalEvent.startDate);
         icalEvent.endDate.addDuration(diff);
         var newEndDate = icalEvent.endDate;
         icalEvent.endDate = newEndDate;
         icalEvent.startDate = newStartDate;
-        
+
+        var rrules = event.getAllProperties('rrule');
+        Logger.log(icalEvent.summary);
+        rrules.forEach(function(r){
+          var vals = r.getValues();
+          vals.forEach(function(v,i){
+            var ct = v.count;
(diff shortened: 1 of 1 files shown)


jonas0b1011001:
> (I have the feeling this could be simplified a lot if the ICAL.RecurExpansion class could provide the total number of occurrences for time-limited recurrences)

I agree, have you though about manually counting the new occurrences with a [RecurIterator](https://mozilla-comm.github.io/ical.js/api/ICAL.RecurIterator.html)?

I also though about counting within the [recurExpansion](https://github.com/derekantrican/GAS-ICS-Sync/blob/d24af0ce0ce9aea0f1f8db55432d4172b61ee8c3/Helpers.gs#L482) but that will not work for multiple rrules.

rufuswilson:
From the specs, it seems that only one rrule can be specified:
_it SHOULD NOT be specified more than once. The recurrence set generated with multiple "RRULE" properties is undefined._

This means that we can simply count directly in the recurExpansion loop, isn't?

jonas0b1011001:
True. Something like this should work

if (newStartDate == null){
  newStartDate = next;
}
if icalEvent.isByCount(){
 newCount ++;
}
else{
  break;
}

jonas0b1011001 (approved):
Nice, looks good to me.

rufuswilson:
Just to check, is there anything missing to merge it in?

jonas0b1011001:
I'm happy with it.
If Derek agrees it is ready to be merged.

derekantrican:
Was busy all weekend. Merging now

PR #243, opened 2021-12-13
Adding some color to the events

leog:
I needed to differentiate between my original events and my imported events through this marvelous script so I looked up that the `colorId` was available and unused so I added it. Hopefully this is also a nice-to-have for other people which is why I wanted to contribute it 👍


colorIds.gs
```diff
@@ -0,0 +1,14 @@
+var colorIds = [
+  "unknown",
```
derekantrican, line 2: Lets make this "default" rather than "unknown"

Code.gs
```diff
@@ -22,10 +22,10 @@
 */
 
 var sourceCalendars = [                // The ics/ical urls that you want to get events from along with their target calendars (list a new row for each mapping of ICS url to Google Calendar)
-                                       // For instance: ["https://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics", "US Holidays"]
-  ["icsUrl1", "targetCalendar1"],
-  ["icsUrl2", "targetCalendar2"],
-  ["icsUrl3", "targetCalendar1"]
+                                       // For instance: ["https://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics", "US Holidays",  "colorId 1 to 11: check colorIds.gs file"]
+  ["icsUrl1", "targetCalendar1", "11"],
```
derekantrican, line 26: I really don't want to make this 3rd parameter required. That will be pretty confusing for new users.

Can we make it so a 3rd parameter to the calendar map (`[targetCalendar, icsURL, color]`) is optional rather than required? Then we could send -1 to `processEvent` if that 3rd parameter is unspecified (ie as the map is currently)

Code.gs
```diff
@@ -151,16 +151,17 @@ function startSync(){
...
     //------------------------ Fetch URL items ------------------------
     var responses = fetchSourceCalendars(sourceCalendarURLs);
     Logger.log("Syncing " + responses.length + " calendars to " + targetCalendarName);
     
     //------------------------ Get target calendar information------------------------
-    var targetCalendar = setupTargetCalendar(targetCalendarName);
+    var targetCalendar = setupTargetCalendar(targetCalendarName, colorId);
     targetCalendarId = targetCalendar.id;
-    Logger.log("Working on calendar: " + targetCalendarId);
+    Logger.log("Working on calendar: " + targetCalendarId + " with colorId: " + colorIds[colorId]);
```
derekantrican, line 164: Similar to above comment, let's only log "with colorId..." if a color is specified

derekantrican (commented):
Let me know your thoughts on my comments

colorIds.gs
jonas0b1011001: For showing the user the available options i would rather link to the official doc [https://developers.google.com/apps-script/reference/calendar/event-color](https://developers.google.com/apps-script/reference/calendar/event-color)

No need for an additional file/manually updated array

leog: Got it, added a reference to it and removed the custom array.

Code.gs
leog: Changed the docs examples to show it is optional.

Code.gs
leog: Made it conditional now.

leog:
Is it simplified enough for you @derekantrican and @jonas0b1011001? I really didn't want to go the unique-color-per-instance way as at least for myself, I need to know which event comes from what calendar. Let me know if you think I need to dig deeper on abstracting the code even more. From my point of view, it is simple enough to get the point without compromising functionality, but it is your call of course 😄

Helpers.gs
```diff
@@ -47,7 +47,7 @@ function condenseCalendarMap(calendarMap){
     if (index > -1)
       result[index][1].push(mapping[0]);
     else
-      result.push([ mapping[1], [ mapping[0] ] ]);
+      result.push([ mapping[1], [ mapping[0] ], mapping[2] ]);
```
derekantrican, line 50: TIL! There's no "index out of bounds" exception in JS

Code.gs
```diff
@@ -23,8 +23,10 @@
 
 var sourceCalendars = [                // The ics/ical urls that you want to get events from along with their target calendars (list a new row for each mapping of ICS url to Google Calendar)
                                        // For instance: ["https://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics", "US Holidays"]
+                                       // Or with colors following mapping https://developers.google.com/apps-script/reference/calendar/event-color, 
+                                       // for instance: ["https://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics", "US Holidays", "11"]
   ["icsUrl1", "targetCalendar1"],
-  ["icsUrl2", "targetCalendar2"],
+  ["icsUrl2", "targetCalendar2", "11"],
```
derekantrican, line 29: I think this is the final request from me - let's remove the "11" in the example array. There's already an example in the comment above and I think that is good enough.

leog: Done 🙂

derekantrican (approved):

jonas0b1011001:
This needs to validate colorId at some point - latest in createEvent().
Right now the code is basically broken if no color is specified in the url array as the API will not accept an event with colorId = null.

derekantrican:
> the API will not accept an event with colorId = null.

I _was_ wondering about that. Figured it was fine, but guess not. @leog can you verify and, if needed, create another PR to do something like `colorId = color ?? "default"` or whatever?

jonas0b1011001:
> the API will not accept an event with colorId = null

Forget that part, it actually does accept null. However i still think it would be smart to validate the colorId.

leog:
Sorry, I just caught your comments post-merge. The value if no colorId is added is actually `undefined`, which seems to work well with the API. And I'm open to add color id validations, but I think it will overcomplicate things a bit with no real benefit other than making sure people selects a valid and desired color, which we just made it totally optional to begin with 🤔

PR #247, opened 2022-01-01
Sync Tasks without dueDate and avoid error

AndBaetz:
This is my first commit ever and I barely know how to even do this and how JS works. I am using NC Tasks and want to sync them to Google, but many of them have no "dueDate", so my idea is to set one at today+30 days (or whatever) if there is none available (could be a var in the Code.gs, but Idk how to). Thank you.


AndBaetz:
Just noticed that I have Tasks which are done in my .ics-File. They have "STATUS:COMPLETED" and they still get added. Guess there is a check missing if they are already completed?

derekantrican:
I appreciate the PR, but I think this is a fairly niche situation. So feel free to keep the change for your own script, but I don't think we'll integrate it into the main project

PR #252, opened 2022-01-18
Make minor updates to instructions

simonkotwicz:


Code.gs
```diff
@@ -6,15 +6,13 @@
 * 1) Make a copy:
 *      New Interface: Go to the project overview icon on the left (looks like this: ⓘ), then click the "copy" icon on the top right (looks like two files on top of each other)
 *      Old Interface: Click in the menu "File" > "Make a copy..." and make a copy to your Google Drive
-* 2) Settings: Change lines 24-47 to be the settings that you want to use
-* 3) Install:
-*      New Interface: Make sure your toolbar says "install" to the right of "Debug", then click "Run"
-*      Old Interface: Click "Run" > "Run function" > "install"
+* 2) Settings: Change lines 22-48 to be the settings that you want to use
```
derekantrican, line 9: Why were the line numbers changed?

derekantrican (commented):
Why are these changes suggested? Users can still access the old interface by using "Use classic editor" in the top right (and currently there's still a number of features that haven't been ported to the new interface, so there's still reason to use the old one)

derekantrican (approved):

simonkotwicz:
Sorry, my bad, I didn't realize the old editor was still available. I updated the PR to just update the line numbers for the personal config and added the new interface instructions for the uninstall function (similar to other places)

PR #276, opened 2022-06-22
Adding the option to anonymize events copied (useful for just blocking the availability)

etairaz:


mattbigger:
In Helpers.gs, line 459 you have 
`if (anonymizezEvents) {`

Isn't the var you created in Code.gs anonymizeEvents without the extra z?

etairaz:
> In Helpers.gs, line 459 you have `if (anonymizezEvents) {`
> 
> Isn't the var you created in Code.gs anonymizeEvents without the extra z?

Yep, sorry about that. I must have hit z accidentally, when I copied from my scripts in google. Fix coming

PR #279, opened 2022-07-11
Event Filtering Functionality

onghaim:
Just a quick PR for filtering out events from the original ical feed, I've been using this to keep meetings on my original calendar while decluttering my personal calendar.


ryancasler:
All the 👍 I have!   Absolutely wonderful addition.  I have been trying to figure out a way to filter out Private Appointments and this worked beautifully.  Thank you!!

PR #280, opened 2022-07-15
On event updated

shewfig:
Small addition to add support for Google Calendar "onEventUpdated" push triggers


derekantrican:
First off - thanks for the PR, but I doubt it'll be merged in.

But, can you explain this a bit more? This is the first time I've ever seen this trigger. But is the intention to run the `startSync` function every time the Google Calendar is updated?

shewfig:
Exactly. The trigger fires from the gCal side when the upstream calendar is
changed, which means the downstream gets updates quicker.

No worries if you don't merge. I sent the PR as a courtesy to let you know
that I appreciate what you've built, and to let you know that there's
additional functionality supported in case you want to revisit your code -
but I also realize that most of us create automation with the hope that we
will never again have to think about it.

With update latency handled by the push trigger, I also added logic in my
deployment to widen the time-based trigger frequency up to 12hr.  Periodic
polling is still a good idea in case the push trigger doesn't fire, but
every few hours is (IMHO) more than sufficient as a redundant sweep.

Unfortunately the update trigger doesn't carry specific info: it is only a
notification that a change happened somewhere, but that makes it perfect to
replace a periodic poll.

Interestingly, MS O365 does a much more thorough job of triggers with their
Power Automate service, but the trade-off is that their only UI is a
click/drag "no-code" web IDE, both for editing and for logging.

Cheers,
-Jim

derekantrican:
Ah, so it sounds like you're using the script to update Google Calendar -> ICS -> this script -> a different Google Calendar? That would make sense with the trigger you're suggesting.

NoSubstitute:
@shewfig - MS Power Automate also sometimes (like most sync services) goes crazy and creates thousands of duplicates of a bunch of events. I haven't seen any cloud based sync service wihtout such glitches. It's painful to clean up.

PR #316, opened 2023-02-21
Color regex

non-bin:
colors events based on their title. I use it for my for my unin calendar, to give events for each class the same color


derekantrican:
Appreciate the PR and desire to contribute, but I think this is too specific to your situation to include in the script for everyone.

Issue #117, opened 2020-04-20
Option to preserve past events that have been removed from source


non-bin:
A calendar I sync from removes events after they have passed, an option (per source calendar?) to cause `processEventCleanup()` to ignore past events would be beneficial.

I still want events in the future to be removed if they are cancelled, so `removeEventsFromCalendar = false` would not work

jonas0b1011001:
Are you looking for the same thing as #109 ?

non-bin:
Yes, thank you. Is this likely to be added as a feature? I would be happy to create a pull request with the changes

derekantrican:
At this point I don't think we'll include it as a standard feature as it's a non-standard situation (ICS source calendar removing events as they've passed, but you want to keep them on your calendar - usually ICS calendars *don't* do that). But so far we've had 2 people with this situation so if we get a lot of people requesting this it would definitely be a good reason to have a built-in feature.

Since your issue is solved, I'm going to go ahead and close this issue.

hubikj:
@derekantrican Sorry for replying to an old closed issue. Unfortunately, this is a standard behavior. I have Outlook 365 calendar at work and it has exactly this behavior - the ICS contains only recent events. Anything older than 1 month is removed, which is really annoying. It would be best to have sync option to not remove anything older than 1 month.

derekantrican:
@hubikj per https://github.com/derekantrican/GAS-ICS-Sync/issues/125 the published time range of Outlook events is 3 months ago to 6 months from now. If you are an admin of your Office 365 account you can change it by following these steps: https://github.com/derekantrican/GAS-ICS-Sync/issues/125#issuecomment-631626145

As far as a change to the script, have you looked at https://github.com/derekantrican/GAS-ICS-Sync/issues/168#issuecomment-734653434?

hubikj:
@derekantrican I am not the admin of our corporate email, so I cannot change this. I used the modification from #168 successfully, but I think it should be a standard part of this script since it is super helpful. I just commented here to say that this behavior is more common than you may think. Thanks a lot!

lyssbjack:
Just popping in to one up this functionality.  I have multiple cals I'm subscribed to that do not keep past events.

Issue #292, opened 2022-11-03
Add option to only resync future events (but keeping the past events)


FoxtrotSierra6829:
My source calendar unfortunately removes past events from more than a month ago. For future reference I want to keep them though. Still I need the ability to sync the removal of future events.

Currently there is an existing option to only sync future events:

`var onlyFutureEvents = false;          // If you turn this to "true", past events will not be synced (this will also removed past events from the target calendar if removeEventsFromCalendar is true)`

As I need to have `removeEventsFromCalendar` set to `true` in order to be able to remove future events, this will also remove past events from the calendar though.

jonas0b1011001:
Have a look at #168 this might be what you are looking for.

FoxtrotSierra6829:
> Have a look at #168 this might be what you are looking for.

This solution seems to be working for me. Can this be integrated into master in a future PR?

derekantrican:
@jonas0b1011001 - while this is definitely a niche request, it seems there are about 5 users that wanted something similar. Maybe worth adding an option for it? Thoughts?

(maybe once https://github.com/derekantrican/GAS-ICS-Sync/discussions/311 is completed)

jonas0b1011001:
There actually is a PR ready to be merged in [https://github.com/derekantrican/GAS-ICS-Sync/pull/317](https://github.com/derekantrican/GAS-ICS-Sync/pull/317).

PR #317, opened 2023-02-21
add option to not delete past events

non-bin:
addresses #117 by adding a toggle, to disable deleting past events

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -636,12 +636,12 @@ function processEventInstance(recEvent){
  * Deletes all events from the target calendar that no longer exist in the source calendars.
  * If onlyFutureEvents is set to true, events that have taken place since the last sync are also removed.
  */
-function processEventCleanup(){
+function processEventCleanup(cleaupPastEvents){
   for (var i = 0; i < calendarEvents.length; i++){
       var currentID = calendarEventsIds[i];
       var feedIndex = icsEventsIds.indexOf(currentID);
       
-      if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
+      if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null && (cleaupPastEvents || calendarEvents[i].start.dateTime > new Date().toISOString())){
         Logger.log("Deleting old event " + currentID);
         callWithBackoff(function(){
           Calendar.Events.remove(targetCalendarId, calendarEvents[i].id);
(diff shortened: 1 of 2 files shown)


Helpers.gs
```diff
@@ -636,12 +636,12 @@ function processEventInstance(recEvent){
  * Deletes all events from the target calendar that no longer exist in the source calendars.
  * If onlyFutureEvents is set to true, events that have taken place since the last sync are also removed.
  */
-function processEventCleanup(){
+function processEventCleanup(cleaupPastEvents){
```
derekantrican, line 639: I don't think you need to pass this through as an argument to the method. I think you can just reference this setting variable in line 644 without this argument.

non-bin: You're right, I've also found that it doesn't keep events properly that have passed on the current day, just a minuite

non-bin:
ok it all works now, but I'm not 100% sure it covers all edge cases. The documentation mentions `start.date` but I think that's just for recurring events which are ignored by this method anyway, but then for `start.dateTime` it says:

> A time zone offset is required unless a time zone is explicitly specified in `timeZone`.

but doesn't mention what will be returned by the API, it seems to always put a timezone in `dateTime` and `timeZone` though.
It's annoying that we can't just use the standard API, that would just return a Date object and remove all these problems.

[https://developers.google.com/calendar/api/v3/reference/events#start.date](https://developers.google.com/calendar/api/v3/reference/events#start.date)
[https://developers.google.com/apps-script/reference/calendar/calendar-event#getStartTime()](https://developers.google.com/apps-script/reference/calendar/calendar-event#getStartTime())

jonas0b1011001:
> The documentation mentions `start.date` but I think that's just for recurring events

start.date is used for all-day events and should be considered aswell.

non-bin:
thanks, how's that?

jonas0b1011001:
That will work!

derekantrican:
Also #292

Helpers.gs
```diff
@@ -641,7 +641,7 @@ function processEventCleanup(){
       var currentID = calendarEventsIds[i];
       var feedIndex = icsEventsIds.indexOf(currentID);
 
-      if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null){
+      if(feedIndex  == -1 && calendarEvents[i].recurringEventId == null && (removePastEventsFromCalendar || new Date(calendarEvents[i].start.dateTime) > new Date() || new Date(calendarEvents[i].start.date) > new Date())){
```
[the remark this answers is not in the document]
non-bin: how's this now? I'm open to suggestions

derekantrican (approved):

PR #319, opened 2023-02-26
Fix organizer handling

jonas0b1011001:


Helpers.gs
```diff
@@ -359,10 +359,20 @@ function createEvent(event, calendarTz){
...
     newEvent.summary = icalEvent.summary;
 
-  if (addOrganizerToTitle && event.hasProperty('organizer')){
-    var organizer = event.getFirstProperty('organizer').getParameter('cn').toString();
-    if (organizer != null)
-      newEvent.summary = organizer + ": " + newEvent.summary;
+  if (event.hasProperty('organizer')){
+    var organizerName = event.getFirstProperty('organizer').getParameter('cn');
+    var organizerMail = event.getFirstProperty('organizer').getParameter('mailto');
+    newEvent.organizer = callWithBackoff(function() {
```
derekantrican, line 365: Is there a reason this needs to be inside `callWithBackoff`? Did you see this fail intermittently?

jonas0b1011001: I have not and i actually don't know if it **can** fail. That's a 'better safe than sorry' solution.

derekantrican (approved):

PR #320, opened 2023-02-28
Upgrade issue templates

derekantrican:
Upgrade issue templates & encourage creation of discussions for support/questions


derekantrican:
May not happen immediately, but after this goes in, we should clean up some of the current issues (keep - if bug/feature, close - if old/fixed, or convert to discussion)

jonas0b1011001 (approved):

PR #349, opened 2023-07-15
Ignoring events feature

juananmgz:
Array of events to be ignored. There you write the summary (title of the event) and it will ignore it to be added, but also delete if already created


jonas0b1011001:
Hi, we already have a [separate branch](https://github.com/derekantrican/GAS-ICS-Sync/tree/event-filtering) with a more versatile filtering function.

PR #358, opened 2023-08-24
Trigger flexibility

Lonestarjeepin:
Adding more flexibility to script trigger to allow for 5, 10, 15, 30, but also any hour from 1-24 with 24 as max trigger.


Lonestarjeepin:
This is my first pull request on Github, so let me know if I messed something up and need to change how I did this!

derekantrican (approved):

PR #373, opened 2023-10-01
fix recurrence exception handling

jonas0b1011001:
- `convertToZone(ICAL.TimezoneService.get('UTC'))` obviously does not work reliably. Replaced with custom calculation/conversion.
- Recurrence exception were added even though `addEventsToCalendar = false`

### DIFF
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -637,16 +637,20 @@ function processEventInstance(recEvent){
   }
 
   if (eventInstanceToPatch !== null && eventInstanceToPatch.length == 1){
-    Logger.log("Updating existing event instance");
-    callWithBackoff(function(){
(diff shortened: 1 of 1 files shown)


derekantrican:
> `convertToZone(ICAL.TimezoneService.get('UTC'))` obviously does not work reliably

Is there an open bug for this on the ICAL.js repo? Or are we just behind?

jonas0b1011001:
We actually are a version behind, the issue however seems to be a known limitation, the readme states:

> The stock ical.js does not register any timezones, due to the additional size it brings. If you'd like to do timezone conversion, and the timezone definitions are not included in the respective ics files, you'll need to use ical.timezones.js or its minified counterpart.

derekantrican (approved):

derekantrican (approved):

PR #400, opened 2023-12-22
6.0 html and json

Lonestarjeepin:
I'm sure you will find ways to run this code more cleanly and likely some things to debug.  Nonetheless, I was able to create the html front-end and json storage in this version.  Each calendar gets run through the startSync individually so that all of the parameters are calendar-specific.  The only variables left in code.gs are the main timer (howFrequnt) and the email variables... I gave up on trying to keep the existing array of arrays so this will have to be a clean break from the old version.  One other thing of note is that this is somewhat based on my syncDelay version so all of the fromGAS entries are different than before so each calendar here can write to the same Target Calendar without conflict.  I'm thinking it would be a good add to look for the old fromGAS=true and just delete all of those entries.  If we don't, there will be duplicates the first time someone runs this code.  Lastly, I updated the  Installation Instructions for this version, but I'm not entirely sure Test Deployments is the "right" way to launch the index.html page, but it works so I just went with it.  It seemed cleaner than messing with Deployments... since each user would have their own deployment anyway.  This version also already includes #372 (it was a building block for this code), #392, and #217 so we can close all of those if this gets merged.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -22,7 +22,7 @@
 *=========================================
 */
 
-var howFrequent = 5;                     // What interval (minutes) to run this script on to check for new events.  Any integer can be used, but will be rounded up to 5, 10, 15, 30 or to the nearest hour after that.. 60, 120, etc. 1440 (24 hours) is the maximum value.  Anything above that will be replaced with 1440.
+var howFrequent = 15;                     // What interval (minutes) to run this script on to check for new events.  Any integer can be used, but will be rounded up to 5, 10, 15, 30 or to the nearest hour after that.. 60, 120, etc. 1440 (24 hours) is the maximum value.  Anything above that will be replaced with 1440.
 var emailSummary = true;                  // Will email you when an event is added/modified/removed to your calendar
 var email = "";                           // OPTIONAL: If "emailSummary" is set to true or you want to receive update notifications, you will need to provide your email address
 var customEmailSubject = "";              // OPTIONAL: If you want to change the email subject, provide a custom one here. Default: "GAS-ICS-Sync Execution Summary"
--- a/Code.gs
+++ b/Code.gs
@@ -6,13 +6,14 @@
 * 1) Make a copy:
 *      New Interface: Go to the project overview icon on the left (looks like this: ⓘ), then click the "copy" icon on the top right (looks like two files on top of each other)
 *      Old Interface: Click in the menu "File" > "Make a copy..." and make a copy to your Google Drive
-* 2) Settings: Change lines 24-50 to be the settings that you want to use
-* 3) Install:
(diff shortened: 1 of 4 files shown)


jonas0b1011001:
> since each user would have their own deployment anyway.

From how i understand deployments (i have no experience at all), the best way would be one 'official' deployment.
If i'm not mistaken you can select to run the deployed code 'as the user' which would allow to save and load all settings from the user properties service. I don't know how/if triggers will work with this.
So in theory the user will only need the html ui and never ever see the code or the code editor.

Lonestarjeepin:
That would be super interesting... and a much easier way to use the web UI.
I'll try to research that and see what the "hosting" would like like. The
embedded version in Apps Script seemed pretty janky.

KW

Lonestarjeepin:
@jonas0b1011001 , try this URL as a "real" web app deployment and see if it works the way you described.  I used the settings of "Execute As: User accessing the web app." and "Who has access: Anyone with Google account."  https://script.google.com/macros/s/AKfycbw1P6zULjJVgElH6MGF7CPs8_m2nZRxjNC851gM1O32mAO4_-VCZ8CCJK98LDJBV_MHUw/exec Not to say I should host this if it does go live, but thought we could at least give this a try in testing.

jonas0b1011001:
I can create/edit/delete calendar entries, but i'm missing a way to actually start/schedule a sync.

Lonestarjeepin:
I left the "Install" in Code.gs.  In hindsight, it would make sense to put
that in the web app somewhere... I was just focused on the calendar
management piece.  I left howFrequent as well as the email summary pieces
in Code.gs.

KW

Lonestarjeepin:
Hey, but at least the calendar piece is working for someone other than me!
I'll take that as a win ;-)

Lonestarjeepin:
@jonas0b1011001 I updated the html to include Install/Uninstall as well as the rest of the App Settings (separate json file called appSettings.json) so now there are zero settings or actions someone would have to take outside of the html.  I updated the same public deployment so you should see the new UI if you use the URL from above.

jonas0b1011001:
The WebApp is creating Triggers, however they are 'empty':
Bildschirmfoto 2023-12-24 um 09 43 30

They get triggered but fail instantly.
Bildschirmfoto 2023-12-24 um 09 42 04

I did some testing earlier and got the same result, could not get it to work. Looks like the way i imagined it to work might be a dead end.

Lonestarjeepin:
Do you think that's strictly a function of the web app being deployed by me instead of on your own profile?  In other words, if you deploy straight from the code, do the triggers work for you?

Lonestarjeepin:
I just confirmed that link does work for me to create a trigger.  So it seems like the single-hosting method might not work... which is unfortunate because it sounded like the preferred method.  I also noticed that when I pushed the last commit, my add/edit calendar functionality is now broken.  I'll troubleshoot and fix now that I'm back from vacation.

Lonestarjeepin:
While this did work "locally", the centrally-hosted html page wouldn't work.  Likely direction is to use config.gs (or similar).

PR #440, opened 2024-05-16
Bump ical.js to v2.0.1

jonas0b1011001:
https://github.com/kewisch/ical.js/releases/tag/v2.0.1


derekantrican:
I think we've made some small tweaks to ical.js in our copy in the past - is that correct? If so, have we made the same changes here?

jonas0b1011001:
Correct, it includes the changes from https://github.com/derekantrican/GAS-ICS-Sync/commit/fae4b1708c3c45352b62e8da26413635360e9c0d and https://github.com/derekantrican/GAS-ICS-Sync/commit/344eb96e89f550e40c62282aa00509de74bc8e4c

derekantrican (approved):

Issue #455, opened 2024-09-11
[Request] Skip Sync When Feed Is Empty


chadn:
### Is your feature request related to a problem? Please describe.

When using this to pull from facebook events, occasionally facebook returns nothing and throws an error in the script.  

When this happens, the script continues as if there were no errors, and deletes every event in the target calendar.
Eventually facebook events return events again, and target calendar is back to normal.   

### Describe the solution you'd like

Create a boolean setting. When true, If no events in the feed, assume a temp problem and do not delete all events in target calendar

iditude:
Hello

I face the same issue. I changed the script following the commit above, but any chance to merge this? 

Airbnb does the same thing and regularly returns an empty feed. 

Thanks

PR #456, opened 2024-09-11
#455 Skip Sync When Feed Is Empty, and update email

chadn:
#455 if `skipIfEmptyFeed == true`, Skip Sync When Feed Is Empty, and change email to include if there were errors.


LaurensdeGraaff (approved):
Great addition. Works good (tested))

iditude:
@chadn Sorry but I think you have a problem when it fails. The variable reportOverallFailure is never moved to true if the feed is empty so you don't see the error happening. I changed to the below and it works. 

`    //------------------------ Fetch URL items ------------------------
    if (responses.length == 0 && skipIfEmptyFeed){
      Logger.log("Skipping this sync iteration due to empty feed (assuming temp problem), not modifying " + targetCalendarName);
      reportOverallFailure = true;
      continue;
    }
    Logger.log("Syncing " + responses.length + " calendars to " + targetCalendarName);`

PR #482, opened 2025-03-01
feat: Allow persistent GCal deletions

jamesvillarrubia:
**Type:** feat

* **What kind of change does this PR introduce?**
  * This PR introduces a new feature for tracking manually deleted events to prevent them from being re-added during future synchronizations.

* **What is the current behavior?**
  * Previously, the application did not track deleted events, which could lead to events being re-added if they were deleted manually by the user within GCal.

* **What is the new behavior?**
  * The application now tracks manually deleted events and via a state-machine. This includes both single and recurring event instances. Events that are identified as manually deleted will not be re-added during the synchronization process.

* **Does this PR introduce a breaking change?**
  * No breaking changes are introduced.

* **Testing Process** 

  1.   Fully delete test calendar.
  2.   Sync calendar.
  3.   Delete two recurring instances of the same event.  Delete one instance of a different recurring event.  Delete one non recurring event.
  4.   Sync again. 
  5.   Confirm events stay deleted.

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -217,7 +217,7 @@ function startSync(){
         }, defaultMaxRetries);
 
       vevents.forEach(function(e){
-        processEvent(e, calendarTz);
+        processEventWithState(e, calendarTz);
       });
 
       Logger.log("Done processing events");
@@ -238,7 +238,8 @@ function startSync(){
     //------------------------ Add Recurring Event Instances ------------------------
     Logger.log("Processing " + recurringEvents.length + " Recurrence Instances!");
     for (var recEvent of recurringEvents){
-      processEventInstance(recEvent);
+      processEventWithState(recEvent, calendarTz, true); // Use the state machine for recurring instances
+
     }
   }
 
--- a/Helpers.gs
+++ b/Helpers.gs
@@ -863,53 +863,92 @@ function createEvent(event, calendarTz){
  *
  * @param {Calendar.Event} recEvent - The event instance to process
  */
-function processEventInstance(recEvent){
-  Logger.log("ID: " + recEvent.extendedProperties.private["id"] + " | Date: "+ recEvent.recurringEventId);
 
+function processEventInstance(recEvent) {
+  const eventId = recEvent.extendedProperties?.private["id"];
+  const recurrenceId = recEvent.recurringEventId;
+  const fullEventId = recurrenceId ? `${eventId}_${recurrenceId}` : eventId;
+  const currentState = getEventState(fullEventId);
+
+  Logger.log("ID: " + recEvent.extendedProperties.private["id"] + " | Date: " + recEvent.recurringEventId);
+
+  // Attempt to find the existing event instance using a unique identifier
+  var eventInstanceToPatch = findEventInstanceToPatch(eventId, recurrenceId);
+
+  // Check the current state of the event instance
+  switch (currentState) {
+    case EVENT_STATE.UNTRACKED:
+      if (eventInstanceToPatch.length === 0) {
+        if (addEventsToCalendar) {
+          Logger.log("No Instance matched, adding as new event!");
+          callWithBackoff(function() {
+            Calendar.Events.insert(recEvent, targetCalendarId);
+          }, defaultMaxRetries);
+        }
+      } else {
+        Logger.log("Event instance already exists, marking as tracked.");
+      }
+      setEventState(fullEventId, EVENT_STATE.TRACKED);
+      break;
+
+    case EVENT_STATE.TRACKED:
+      if (eventInstanceToPatch.length === 0) {
+        Logger.log("Tracked event instance not found, marking as potentially deleted.");
+        setEventState(fullEventId, EVENT_STATE.POTENTIALLY_DELETED);
+      } else if (modifyExistingEvents) {
+        Logger.log("Updating existing event instance");
+        callWithBackoff(function() {
+          Calendar.Events.update(recEvent, targetCalendarId, eventInstanceToPatch[0].id);
+        }, defaultMaxRetries);
+      }
+      break;
+
+    case EVENT_STATE.POTENTIALLY_DELETED:
+      if (eventInstanceToPatch.length === 0) {
+        Logger.log("Event instance still not found, marking as manually deleted.");
+        setEventState(fullEventId, EVENT_STATE.MANUALLY_DELETED);
+      } else {
+        Logger.log("Event instance found, marking as tracked.");
+        setEventState(fullEventId, EVENT_STATE.TRACKED);
+      }
+      break;
+
+    case EVENT_STATE.MANUALLY_DELETED:
+      Logger.log("Skipping manually deleted event instance " + fullEventId);
+      break;
+
+    case EVENT_STATE.REMOVED:
+      if (icsEventsIds.includes(fullEventId)) {
+        Logger.log("Re-adding removed event instance " + fullEventId);
+        callWithBackoff(function() {
+          Calendar.Events.insert(recEvent, targetCalendarId);
+        }, defaultMaxRetries);
+        setEventState(fullEventId, EVENT_STATE.TRACKED);
+      }
+      break;
+  }
+}
+
+
+/**
+ * Finds an existing event instance in the target calendar that matches the given fullEventId.
+ *
+ * @param {string} fullEventId - The full event ID to search for, in the format "eventId_recurrenceId"
+ * @return {Array.<Calendar.Event>} An array with a single element if a matching event instance is found, or an empty array if no match is found.
+ */
+function findEventInstanceToPatch(recEvent) {
   var eventInstanceToPatch = callWithBackoff(function(){
     return Calendar.Events.list(targetCalendarId,
       { singleEvents : true,
         privateExtendedProperty : "fromGAS=true",
-        privateExtendedProperty : "rec-id=" + recEvent.extendedProperties.private["id"] + "_" + recEvent.recurringEventId
+        privateExtendedProperty : "rec-id=" + recEvent?.extendedProperties?.private["id"] + "_" + recEvent?.recurringEventId
       }).items;
   }, defaultMaxRetries);
 
-  if (eventInstanceToPatch == null || eventInstanceToPatch.length == 0){
-    if (recEvent.recurringEventId.length == 10){
-      recEvent.recurringEventId += "T00:00:00Z";
-    }
-    else if (recEvent.recurringEventId.substr(-1) !== "Z"){
-      recEvent.recurringEventId += "Z";
-    }
-    eventInstanceToPatch = callWithBackoff(function(){
-       return Calendar.Events.list(targetCalendarId,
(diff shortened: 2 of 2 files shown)


jonas0b1011001:
Thank you very much for contributing.

Upon first inspection, the code does not distinguish between
_events that have previously been added but are no longer part of the target calendar_
and 
_events that are newly added to the source and are to be synced for the first time._

So a freshly set up script would never add any events cause the script treats all events as having been manually deleted.

Am i missing something?

jamesvillarrubia:
@jonas0b1011001 Oof, yeah, refreshed this morning and ran into that. Came here this mornign to convert back to draft and saw your comment. So thanks for the speedy and thorough review! 

Not sure why my original manual tests didn't catch this.  I'll come back with a revision.

jamesvillarrubia:
@jonas0b1011001 Totally revised the approach.  Now status (from untracked, tracked, to deleted) is managed via a simple state machine.  States are stored in memory.

I also updated my test process (see PR desc.) to fully wipe the test calendar and deleted multiple different types of events.  

Let me know if you find any other issues or if there's something you think should be added.

jonas0b1011001:
Great, will take a look on the weekend!

Lonestarjeepin:
> @jonas0b1011001 Totally revised the approach.  Now status (from untracked, tracked, to deleted) is managed via a simple state machine.  States are stored in memory.
> 
> I also updated my test process (see PR desc.) to fully wipe the test calendar and deleted multiple different types of events.  
> 
> Let me know if you find any other issues or if there's something you think should be added.

I've sometimes wished I had this feature, but from a user standpoint is there any way to get an event back after deletion? Or what if the source event changes, would it then repopulate? I'm not a good enough programmer to review your code so maybe you've already accounted for this!?! Just some thoughts from a user. Thanks!

jamesvillarrubia:
@Lonestarjeepin they can add a duplicate event in the event source and it'll add back in since the originating ID will be different. Other than that, I'm open to ideas.  If the source event is just updated, I don't believe it will not repopulate since the ID doesn't change.  Perhaps there's an edge case for an event being deleted (for a conflict), but then moved... on the downstream GCal I think it would still be deleted. Not sure how to handle that.

jonas0b1011001:
> I've sometimes wished I had this feature, but from a user standpoint is there any way to get an event back after deletion? Or what if the source event changes, would it then repopulate?

Good questions, just some quick thoughts on it (still have not had a closer look at the new code):

---

1. >  is there any way to get an event back after deletion

Google Calendar has a [trash bin](https://calendar.google.com/calendar/u/0/r/trash) were deleted events sit for 30 days. The user can restore events from there.
`manually_deleted` events that reappear in `calendarEvents` would have to transition back to `tracked`.

---

2. > what if the source event changes, would it then repopulate?

First we would need a way to track that it changed. I guess that would only be possible if we not only track the state of an event but also it's `extendedProperties.private["MD5"]`.
Once we know that it changed, we can
- silently readd the event
- notify the user that a deleted event was updated
- readd and notify
- do nothing

Helpers.gs
```diff
@@ -1350,3 +1389,141 @@ function checkForUpdate(){
...
+      } else {
+        processEvent(event, calendarTz);
+      }
+      setEventState(fullEventId, EVENT_STATE.TRACKED);
+      break;
+
+    case EVENT_STATE.TRACKED:
+      // Check if the event is still in the source
+      if (!icsEventsIds.includes(fullEventId)) {
+        setEventState(fullEventId, EVENT_STATE.REMOVED);
```
jonas0b1011001, line 1444: This code is unreachable, we are iterating over all source events.
`Removed` would have to be set in the cleanup function where we check for events in the target calendar that are not in the source file.

Helpers.gs
```diff
@@ -1350,3 +1389,141 @@ function checkForUpdate(){
...
+      break;
+
+    case EVENT_STATE.MANUALLY_DELETED:
+      // Do nothing, as the event should not be re-added
+      Logger.log("Skipping manually deleted event " + fullEventId);
+      break;
+
+    case EVENT_STATE.REMOVED:
+      // If the event reappears in the source, re-add it
+      if (icsEventsIds.includes(fullEventId)) {
```
jonas0b1011001, line 1475: Unneccessary if statement, at this point we know it reappeared.

Helpers.gs
```diff
@@ -1350,3 +1389,141 @@ function checkForUpdate(){
...
+        if (isRecurringInstance) {
+          processEventInstance(event);
+        } else {
+          processEvent(event, calendarTz);
+        }
+      }
+      break;
+
+    case EVENT_STATE.MANUALLY_DELETED:
+      // Do nothing, as the event should not be re-added
```
jonas0b1011001, line 1469:
```
//Allow recovering manually deleted events
if (calendarEventsIds.includes(fullEventId)) {
  setEventState(fullEventId, EVENT_STATE.TRACKED);
  if (isRecurringInstance) {
    processEventInstance(event);
  } else {
    processEvent(event, calendarTz);
  }
}
```

jonas0b1011001, line 1469: This would also be the place to check if the manually deleted event was updated.

We would have to
- save to latest hash in processEvent/processEventInstance to know which hash the event had when deleted
- calculate the current hash of the event in the source
```
event.removeProperty('dtstamp');
let digest = Utilities.computeDigest(Utilities.DigestAlgorithm.MD5, event.toString(), Utilities.Charset.UTF_8).toString();
```
- compare saved and current hash
- decide what to do in case the hash changed. readd yes/no, notify the user yes/no

Helpers.gs
```diff
@@ -1350,3 +1389,141 @@ function checkForUpdate(){
...
+      // First-time addition
+      if (isRecurringInstance) {
+        processEventInstance(event);
+      } else {
+        processEvent(event, calendarTz);
+      }
+      setEventState(fullEventId, EVENT_STATE.TRACKED);
+      break;
+
+    case EVENT_STATE.TRACKED:
```
jonas0b1011001, line 1441: How do non-recurring events transition from `TRACKED` to `MANUALLY_DELETED`?

You are missing
```
if (!calendarEventsIds.includes(fullEventId)) {
  setEventState(fullEventId, EVENT_STATE.MANUALLY_DELETED);
}
```

Helpers.gs
```diff
@@ -863,54 +863,93 @@ function createEvent(event, calendarTz){
...
+}
+
+
+/**
+ * Finds an existing event instance in the target calendar that matches the given fullEventId.
+ *
+ * @param {string} fullEventId - The full event ID to search for, in the format "eventId_recurrenceId"
+ * @return {Array.<Calendar.Event>} An array with a single element if a matching event instance is found, or an empty array if no match is found.
+ */
+function findEventInstanceToPatch(recEvent) {
```
jonas0b1011001, line 939: This neither matches the doc nor the function call in Line 876.

jonas0b1011001:
I did some testing, for me the current version fails the test procedure you described in your first comment. Normal events are getting readded on every run.
This also reintroduces bugs with general recurrence handling, for example:
create a recurring event, rename only the **first** instance of the series. The instance gets duplicated on first sync.

I'll need more time for a full review/test.

jamesvillarrubia:
Hmmmm... is there a public ics calendar you are using for tests?  Maybe it's specific to the source calendar I'm using. 

I've got calendars from my work outlook, but it doesn't give me much flexibility on test cases. Perhaps there's an ics file somewhere with a broad enough set of test events that we can both work from?

I'll step through the code again to see if there's a logic gap I missed. 

Also any consideration for adding a test suite for the repo?

jonas0b1011001:
> Hmmmm... is there a public ics calendar you are using for tests? Maybe it's specific to the source calendar I'm using.
> 

I was using [this simple ics](https://pastebin.com/raw/HATFJa1r).
`Single` will be readded on each run.
`rec2` and `rec10` get improperly synced because `case EVENT_STATE.UNTRACKED:` (L880) is unreachable, if i fix this (`setEventState(fullEventId, EVENT_STATE.TRACKED);` in L1438 only if recurrenceId is null) i get duplicates.

> I'll step through the code again to see if there's a logic gap I missed.
 
Do you see my comments on the code or are they invisible unless i finish my review?

> Also any consideration for adding a test suite for the repo?

Nothing planned atm.

jamesvillarrubia:
I don't think I can see the comments until you close the review.  Obviously close without approval and I'll debug some more and push some updates.  Thanks for working with me on this1

Code.gs
```diff
@@ -238,7 +238,8 @@ function startSync(){
     //------------------------ Add Recurring Event Instances ------------------------
     Logger.log("Processing " + recurringEvents.length + " Recurrence Instances!");
     for (var recEvent of recurringEvents){
-      processEventInstance(recEvent);
+      processEventWithState(recEvent, calendarTz, true); // Use the state machine for recurring instances
```
jonas0b1011001, line 241: Is there a need for instances to go through the state machine(s) 3 times?

Helpers.gs
```diff
@@ -863,54 +863,93 @@ function createEvent(event, calendarTz){
...
+  const currentState = getEventState(fullEventId);
+
+  Logger.log("ID: " + recEvent.extendedProperties.private["id"] + " | Date: " + recEvent.recurringEventId);
+
+  // Attempt to find the existing event instance using a unique identifier
+  var eventInstanceToPatch = findEventInstanceToPatch(eventId, recurrenceId);
+
+  // Check the current state of the event instance
+  switch (currentState) {
+    case EVENT_STATE.UNTRACKED:
```
jonas0b1011001, line 880: Unreachable - all instances will be set to `TRACKED` the first time they enter the statemachine in processEventWithState().

Helpers.gs
jonas0b1011001, line 939: This function is incomplete and will only ever match instances that where patched by the script (`privateExtendedProperty : "rec-id="`), however we need this to be able to match 'all' instances, i.e. instances that are created by google from a recurring event (recurrence rules are 'expanded' by google and not by the script)

Need to test:
```
function findEventInstanceToPatch(recEvent) {

  if (recEvent.recurringEventId.length > 10 && recEvent.recurringEventId.substr(-1) !== "Z"){
    recEvent.recurringEventId += "Z";
  }
  eventInstanceToPatch = callWithBackoff(function(){
      return Calendar.Events.list(targetCalendarId,
      { singleEvents : true,
        orderBy : "startTime",
        timeZone: "etc/GMT",
        privateExtendedProperty : "fromGAS=true",
        privateExtendedProperty : "id=" + recEvent.extendedProperties.private["id"]
      }).items;
  }, defaultMaxRetries);

  // If the event instance is not found, it may be deleted.

  return eventInstanceToPatch.filter(function(e){
      return ((e.originalStartTime.dateTime || e.originalStartTime.date) == recEvent.recurringEventId);
    });
}
```

jonas0b1011001 (changes requested):
Will need to take a closer look at the recurrence part of it as i'm 100% sure it will cause issues at some point (as it does with **every** feature of the script)

Issue #315, opened 2023-02-20
Script stops syncing calendar events about one week ago


alexnb98:
About last week ago the script stopped syncing my events from my outlook calendar to google calendar. Before that it worked perfectly for 1 and half years (thanks for that!). 

<img width="1341" alt="image" src="https://user-images.githubusercontent.com/35068342/220147704-6d4ae7a5-34fa-47fc-aab1-ab946c2a8fc8.png">

Did something change in the API? What can I do?

IssueHuntBot:
[@alexnb98](https://issuehunt.io/u/alexnb98) has funded $10.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/136078981/issues/315) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/136078981/issues/315).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

derekantrican:
@alexnb98 - is this consistently failing? Can you share your ICS file?

alexnb98:
Yes, it's consistenly failing since 2 weeks. Here is the ICS file, I removed all private information: https://pastebin.com/24DNY3yL

alexnb98:
Hello, how can I find the event id? When I copy the id in the log and paste it into the google calender ics. I find no results?

<img width="1320" alt="image" src="https://user-images.githubusercontent.com/35068342/221500343-63497494-acc7-4ce8-a219-6aa57251ce16.png">

jonas0b1011001:
My bad, make the following change in helpers gs [Line229](https://github.com/derekantrican/GAS-ICS-Sync/blob/e0924b9ea4ca657f02b42588674efbe1b2a7f80f/Helpers.gs#L229)
`Logger.log("Updating existing event " + calendarEvents[index].id);`

Run the script again and you should get a different id in the log.

alexnb98:
Perfect, this is the event that fails. It's an event that repeats and has no end date:

```
UID:nh9ogo5cmk8jm9mdguoodbg5vk_R20230213T110000@google.com
CLASS:PUBLIC
CREATED:20220429T083454Z
DESCRIPTION:\n
LAST-MODIFIED:20230211T101810Z
LOCATION:
SEQUENCE:205
[...]

PR #522, opened 2026-05-11
Skip cancelled existing calendar events

xd-est19xx:
## Summary\n- filter cancelled existing Google Calendar events before building the ID lookup used for updates\n- log how many cancelled events were skipped when this edge case occurs\n\n## Why\nGoogle Calendar can still return cancelled recurring-event instances in some edge cases even when listing with showDeleted=false. If one of those cancelled events matches an incoming ICS event, updating it can fail with `Invalid start time`.\n\nFixes #315\n\n## Validation\n- `cp Code.gs /tmp/gas-ics-Code.js && node --check /tmp/gas-ics-Code.js`

### DIFF
--- a/Code.gs
+++ b/Code.gs
@@ -195,6 +195,13 @@ function startSync(){
         if (eventList != null)
           calendarEvents = [].concat(calendarEvents, eventList.items);
       }
+      var calendarEventsCount = calendarEvents.length;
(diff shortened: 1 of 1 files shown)


xd-est19xx:
Closing this because I noticed #520 already proposes the same cancelled-event filtering fix for #315. No need to duplicate that work.

